PR #65, opened 2018-04-17
Add link to ElasticSearch README

black-adder:
README has info on purging old indexes and other implementation details.

Signed-off-by: Won Jun Jang <wjang@uber.com>

### DIFF
--- a/content/docs/deployment.md
+++ b/content/docs/deployment.md
@@ -166,6 +166,8 @@ ElasticSearch does not require initialization other than
 [installing and running ElasticSearch](https://www.elastic.co/downloads/elasticsearch).
 Once it is running, pass the correct configuration values to the Jaeger collector and query service.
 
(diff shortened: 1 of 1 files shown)


content/docs/deployment.md
```diff
@@ -166,6 +166,8 @@ ElasticSearch does not require initialization other than
 [installing and running ElasticSearch](https://www.elastic.co/downloads/elasticsearch).
 Once it is running, pass the correct configuration values to the Jaeger collector and query service.
 
+See the [README](https://github.com/jaegertracing/jaeger/tree/master/plugin/storage/es) for an in-depth overview of how Jaeger uses ElasticSearch for storage.
```
yurishkuro: nit: link to README.md file directly, because otherwise it leads to source code view

yurishkuro: NB: the readme says "This provides a plugin to use Jaeger with ElasticSearch. This currently supports ElasticSearch 5.x." - we need to mention 6.x too

PR #68, opened 2018-05-02
Add link to April Newsletter

yurishkuro:
Temporary solution for #69


yurishkuro:
@lucperkins this is just a quick fix before KubeCon, do you think there is a better place for this?

PR #82, opened 2018-05-25
Update C# client features

cwe1ss:
The new v0.0.10 release of the C# client now supports almost the same features as Java.

I couldn't run this locally so I couldn't test it. Hope it's fine though.

(resolves https://github.com/jaegertracing/jaeger-client-csharp/issues/51)

### DIFF
--- a/data/clients.yaml
+++ b/data/clients.yaml
@@ -291,38 +291,38 @@ clients:
     repo: jaegertracing/jaeger-client-csharp
     features:
       report_jaeger_thrift_udp: yes
-      report_jaeger_thrift_http:
+      report_jaeger_thrift_http: yes
       report_zipkin_thrift_http: no
       propagation_uber: yes
-      propagation_b3: no
+      propagation_b3: yes
       propagation_w3c: coming
-      jaeger_debug_id:
-      jaeger_baggage:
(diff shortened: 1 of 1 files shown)


yurishkuro:
Tip: if you can't test locally you can check the preview in the PR

data/clients.yaml
```diff
@@ -291,38 +291,38 @@ clients:
...
-      env_JAEGER_REPORTER_FLUSH_INTERVAL: no
-      env_JAEGER_SAMPLER_TYPE: no
-      env_JAEGER_SAMPLER_PARAM: no
-      env_JAEGER_SAMPLER_MANAGER_HOST_PORT: no
-      env_TRACER_TAGS: no
+      configuration: yes
+      config_tracer_tags: yes
+      remote_sampler: yes
+      adaptive_sampler: yes
+      baggage_restriction: coming
```
Falco20019: Restrictions are completely implemented now, right?

Falco20019: @cwe1ss Friendly ping

yurishkuro:
I a going to merge this, minor corrections can be done separately. Baggage restrictions are not well supported from the backend side anyway - @black-adder is this something we can add as a static config?

PR #91, opened 2018-06-15
Add instructions to run HotROD from Docker image

### LINKED ISSUE #60: Replace HotROD running instructions with Docker image steps
https://www.jaegertracing.io/docs/getting-started/#running

chandresh-pancholi:
Signed-off-by: chandresh-pancholi <chandreshpancholi007@gmail.com>

- Resolves #60 

- Replacing traditional way of running hotrod example to docker way

### DIFF
--- a/content/docs/getting-started.md
+++ b/content/docs/getting-started.md
@@ -62,6 +62,17 @@ It can be run standalone, but requires Jaeger backend to view the
 
 #### Running
 
+###### From Source
+```bash
+mkdir -p $GOPATH/src/github.com/jaegertracing
+cd $GOPATH/src/github.com/jaegertracing
+git clone git@github.com:jaegertracing/jaeger.git jaeger
+cd jaeger
+make install
+cd examples/hotrod
+go run ./main.go all
+```
+##### From docker
 ```bash
 $ docker run --rm -it \
   --link jaeger \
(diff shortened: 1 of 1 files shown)


content/docs/getting-started.md
```diff
@@ -63,13 +63,11 @@ It can be run standalone, but requires Jaeger backend to view the
...
 
 ```bash
-mkdir -p $GOPATH/src/github.com/jaegertracing
-cd $GOPATH/src/github.com/jaegertracing
-git clone git@github.com:jaegertracing/jaeger.git jaeger
-cd jaeger
-make install
-cd examples/hotrod
-go run ./main.go all
+$ docker run --rm -it \
```
yurishkuro: I think it's good to add a docker command as an easier way to run, but we should keep the original instructions as well, because the point of this application, as described in the blog post, is to find and solve performance issues, which requires running from source.

chandresh-pancholi: I will include both the steps to run Hotrod example

content/docs/getting-started.md
```diff
@@ -62,6 +62,7 @@ It can be run standalone, but requires Jaeger backend to view the
 
 #### Running
 
+###### From Source
```
yurishkuro: This should be 5 hashes

yurishkuro:
Thanks!

PR #116, opened 2018-07-26
Fix Nav UI's issue

### LINKED ISSUE #115: Fix small UI in Nav
Hi,
When I go to Features page in Docs in resolution 1366x768, the page displayed like this.

screenshot from 2018-07-26 10-03-06

So I want to fix that by adding an css attribute into `.page--docs .nav`

And result:

screenshot from 2018-07-26 10-03-20

I will create a pull request soon.

chidokun:
This PR add an css attribute to `.page--docs .nav` to break the nave item into new line in case the text is long. It will make web page display more reasonable in 1366x768 screen resolution.

Resolves #115

### DIFF
--- a/themes/jaeger-docs/assets/sass/pages.sass
+++ b/themes/jaeger-docs/assets/sass/pages.sass
@@ -25,6 +25,7 @@
     /* Left-hand documentation nav */
     .nav
       position: fixed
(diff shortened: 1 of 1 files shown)


yurishkuro:
Thanks for the fix!

PR #119, opened 2018-08-05
Reinstate Download button

### LINKED ISSUE #114: Reinstate Download button
Singe we now create downloadable binaries on GitHub releases, we can re-enable the Download button on the home page:

https://github.com/jaegertracing/documentation/blob/a00fb20258b33e0dc74717a3442e847c700fd1da/themes/jaeger-docs/layouts/partials/home/hero.html#L25

We can either point it directly to GitHub releases, or better to a new Download page that will add a bit of explanation, like mention that Docker and binaries are the two alternatives.

cc @PikBot

PikBot:
Signed-off-by: Prakriti <prakritibansal98@gmail.com>

- resolves #114 

- Add download docs 
- Reinstate download button

### DIFF
--- a/content/docs/download.md
+++ b/content/docs/download.md
@@ -1,0 +1,11 @@
+---
+title: Download
+weight: 13
+menu: docs
+---
+
+## Binaries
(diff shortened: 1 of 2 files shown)


yurishkuro:
Thanks!

Issue #90, opened 2018-06-15
Document Kafka storage usage


jpkrohling:
## Requirement - what kind of business use case are you trying to solve?

Use the new Kafka storage.

## Problem - what in Jaeger blocks you from solving the requirement?

Lack of documentation about it.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Document it:

* How to start the components (standalone/query/collector) with the new storage
* What configuration options there are
* Gotchas (write-only!)
* Bonus points: how to start Kafka for development purposes (Docker?)
* Missing bits (the component that ingests data and writes to Cassandra/Elasticsearch)

PikBot:
Are there some resources available as a starting point? I don't have much knowledge myself about Kafka storage but I would like to help out with this issue.

yurishkuro:
cc @vprithvi

vprithvi:
@PikBot There aren't any resources apart from the code at this juncture. I'm familiar with Kafka and can write a skeleton doc that we can iterate on.

PikBot:
@vprithvi that would be helpful. 

Also can you share the link to the code? I can go through it and try to understand it from a technical point of view.

vprithvi:
@PikBot The code for emitting to Kafka is at  https://github.com/jaegertracing/jaeger/tree/master/plugin/storage/kafka 

We are working on releasing the code for consuming from Kafka and writing to Cassandra, it's being tracked as part of https://github.com/jaegertracing/jaeger/issues/929

PikBot:
@vprithvi thanks! I'll try my best to get it documented. An initial skeleton doc from your side would be nice!

PR #129, opened 2018-08-27
Updating the documentation to add Kafka

ledor473:
Trying to resolves https://github.com/jaegertracing/documentation/issues/90, but there might be some stuff missing

- Kafka section
- Configuration section to each storage types to know what is the minimal config and how to find the complete list of options
- The Jaeger version needed to use ES and Kafka


content/docs/deployment.md
```diff
@@ -208,6 +239,34 @@ Shards and replicas are some configuration values to take special attention to,
 index creation. [This article](https://qbox.io/blog/optimizing-elasticsearch-how-many-shards-per-index) goes into
 more information about choosing how many shards should be chosen for optimization.
 
+### Kafka
+Supported in Jaeger since 1.6.0
+Supported Kafka versions: 0.8+
+
+Considering the Kafka storage backend implementation only supports writing, you will need to use the multiple storage types feature.
```
yurishkuro: Add: Writing to Kafka is particularly useful for building post-processing data pipelines.

ledor473:
I've rephrased it, added your comment and included a mention of the upcoming "ingester" mechanism/component (haven't look at that part yet)

content/docs/deployment.md
```diff
@@ -208,6 +239,37 @@ Shards and replicas are some configuration values to take special attention to,
 index creation. [This article](https://qbox.io/blog/optimizing-elasticsearch-how-many-shards-per-index) goes into
 more information about choosing how many shards should be chosen for optimization.
 
+### Kafka
+Supported in Jaeger since 1.6.0
+Supported Kafka versions: 0.8+
```
yurishkuro: this looks like this:
image

perhaps use bullets

content/docs/deployment.md
```diff
@@ -119,7 +119,9 @@ Port  | Protocol | Function
```
yurishkuro: suggest `## Storage Backend` -> `## Storage Backends` (plural)

content/docs/deployment.md
```diff
@@ -208,6 +239,37 @@ Shards and replicas are some configuration values to take special attention to,
 index creation. [This article](https://qbox.io/blog/optimizing-elasticsearch-how-many-shards-per-index) goes into
 more information about choosing how many shards should be chosen for optimization.
 
+### Kafka
+Supported in Jaeger since 1.6.0
+Supported Kafka versions: 0.8+
+
+Currently, the Kafka storage backend implementation only supports writing data, this means you will need to use another one (with the multiple storage types feature) to be able to view the traces in the Query component.
```
vprithvi: >Currently, the Kafka storage backend implementation only supports writing data

This is no longer true, but we don't yet have documentation for the `ingester`. 

>another one (with the multiple storage types feature) to be able to view the traces in the Query component.

I'd replace one with storage

content/docs/deployment.md
```diff
@@ -208,6 +239,37 @@ Shards and replicas are some configuration values to take special attention to,
...
+  jaegertracing/jaeger-collector
+```
+
+##### All options
+To view the full list of configuration options, you can run the following command:
+```sh
+docker run -e SPAN_STORAGE_TYPE=kafka jaegertracing/jaeger-collector:1.6 --help
+```
+
+#### Topic & partition
```
vprithvi: s/partition/partitions

content/docs/deployment.md
```diff
@@ -208,6 +239,37 @@ Shards and replicas are some configuration values to take special attention to,
...
+Unless your Kafka cluster is configured to automatically create topics, you will need to create it ahead of time. You can refer to [the Kafka quickstart documentation](https://kafka.apache.org/documentation/#quickstart_createtopic) to learn how.
+
+You can also find more information about topics and partitions in the [official documentation](https://kafka.apache.org/documentation/#intro_topics).
```
vprithvi: Perhaps we should link to https://www.confluent.io/blog/how-to-choose-the-number-of-topicspartitions-in-a-kafka-cluster/ for selecting a reasonable number of partitions when creating the topic?

I'm a bit conflicted on whether we should even include partitions in this section of the documentation because someone operating Kafka should be pretty familiar with the concept.

yurishkuro: I think every extra bit helps, +1 to link to a blog

yurishkuro: especially if you (@vprithvi) found it useful when connecting Jaeger to Kafka

content/docs/deployment.md
ledor473: @vprithvi regarding the ingester, afaik it's not even released? It would probably be bundled as part of 1.7 so I can mention that instead of "in a later release"

content/docs/deployment.md
ledor473: 👍

yurishkuro:
Thanks!

PR #134, opened 2018-08-30
fix link to client libraries doc

The-Alchemist:


yurishkuro:
the links are correct: https://www.jaegertracing.io/docs/getting-started/

PR #156, opened 2018-09-28
Add 1.7 release

yurishkuro:


yurishkuro:
no change except for the config, just a straight copy of 1.6

we can make additional changes in 1.7 dir now, since some of the new features are not reflected.

PR #162, opened 2018-11-02
October 2018 Newsletter

jpkrohling:


.gitignore
```diff
@@ -2,3 +2,4 @@ public/
 .DS_Store
 .bundle/
 vendor/bundle/
+resources/_gen
```
yurishkuro: we currently check in these due to issue with Hugo on Netlify

jpkrohling: Got it. PR updated!

PR #171, opened 2018-11-15
Add archive traces UI configuration

pavolloffay:

Resolves: https://github.com/jaegertracing/jaeger-ui/issues/7

### DIFF
--- a/content/docs/1.6/deployment.md
+++ b/content/docs/1.6/deployment.md
@@ -314,6 +314,7 @@ An example configuration file:
     "dagMaxNumServices": 200,
     "menuEnabled": true
   },
+  "archiveEnabled": true,
   "tracking": {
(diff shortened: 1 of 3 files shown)


pavolloffay:
you can raise it in the UI repo

pavolloffay:
Thinking about this, Not sure if there are more conf props for the archive. I think dependencies is an object because there are two properties.

pavolloffay:
+1 will clarify

tiffon:
@pavolloffay You're right, the `archiveEnabled` is a top-level property because it's a single isolated value. 

I commented with trade-offs to both approaches in the UI ticket: https://github.com/jaegertracing/jaeger-ui/issues/277#issuecomment-442321167

What do you think?

PR #176, opened 2018-11-19
Update instructions on making a new release

### LINKED ISSUE #174: Rethink "next version" docs location
We've run into this in the last release where all files were copied from 1.7 to 1.8 dir with simultaneous changes for the new features, which makes it difficult to review and track.

I would like to propose a different mechanism:
  * have a folder called `next-release` or something like that
    * changes for new features are made in that folder
    * there should be a way to see those docs rendered both in the Netlify PR builds and locally
    * on the main production site we can treat the `next-release` folder in one of two ways
      * never show on the main site
      * show if accessed by direct URL, but never show in the list of versions
  * when we cut a new release, it is just a straight copy from `next-release` to `major.minor` dir, without any additional changes (meaning we don't want to have special labels in the `next-release` files metadata indicating that they should be treated differently, the logic should only be based on the directory name).

If it is easier to implement, we don't have to keep these files in the `content/docs/next-release`, where they may be picked up by the version listing logic, we could pull them to a level higher, like `content/docs-next-release`. With that we may not even need any template changes at all.

cc @lucperkins @pavolloffay @jpkrohling

yurishkuro:
Resolves #174

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -24,7 +24,6 @@ home = [ "HTML", "REDIRECTS" ]
   mediumHandle   = "jaegertracing"
   gitterHandle   = "jaegertracing"
   openGraphImage = "img/jaeger-icon-color.png"
-  newsletterTitle = "October 2018 project update"
   maxBlogPosts   = 100
 
   latest         = "1.8"
--- a/README.md
+++ b/README.md
@@ -30,19 +30,14 @@ We strongly encourage you to contribute to this site! For more information, see 
 
 ### Publishing new Jaeger version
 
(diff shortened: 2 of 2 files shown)


config.toml
```diff
@@ -24,7 +24,6 @@ home = [ "HTML", "REDIRECTS" ]
   mediumHandle   = "jaegertracing"
   gitterHandle   = "jaegertracing"
   openGraphImage = "img/jaeger-icon-color.png"
-  newsletterTitle = "October 2018 project update"
```
jpkrohling: Is this change intentional?

yurishkuro: Yes, cleaning up a leftover from earlier change that gets this title directly from medium

PR #209, opened 2019-01-23
Enrich Zipkin compatible endpoint documentation

### LINKED ISSUE #208: Enrich Zipkin compatible endpoint documentation in get-started for v1.9

User should be able to understand clearly what format is used in which endpoint from [documentation on Zipkin compatible endpoints](https://www.jaegertracing.io/docs/1.9/getting-started/#migrating-from-zipkin).

Existing documentation might be misleading. Currently, it can be easily misunderstood by people who are not familiar with API versions of Zipkin. 
This has not really huge impact. Nevertheless, would be better to have this documentation improved.

Current state is 
> Collector service exposes Zipkin compatible REST API `/api/v1/spans` and `/api/v2/spans` for both JSON and thrift encoding.

I propose to change it to
> Collector service exposes Zipkin compatible REST API `/api/v1/spans` and `/api/v2/spans` for both
Thrift and JSON formats correspondingly.

d1mnewz:
Described in #208 

Obvious. Described in #208


content/docs/1.9/getting-started.md
```diff
@@ -120,7 +120,7 @@ Then navigate to `http://localhost:8080`.
 ## Migrating from Zipkin
 
 Collector service exposes Zipkin compatible REST API `/api/v1/spans` and `/api/v2/spans` for both
-JSON and thrift encoding.
+Thrift and JSON formats correspondingly.
```
yurishkuro: that's not accurate. The v1 endpoint accepts both JSON and Thrift, while v2 only accepts JSON (since there's no such thing as Thrift v2 in Zipkin).

https://github.com/jaegertracing/jaeger/blob/d23d18f43147431e03eedd10b82b603388964f91/cmd/collector/app/zipkin/http_handler.go#L91

pavolloffay: hi @d1mnewz could you please update the PR?

d1mnewz: Sure! Will do that right away.

yurishkuro:
this PR changes the docs for the old release, I suggest making change in the `next-release` folder instead (this change is not that critical to apply retroactively).

Please make sure commits are signed.

d1mnewz:
> this PR changes the docs for the old release, I suggest making change in the `next-release` folder instead (this change is not that critical to apply retroactively).
> 
> Please make sure commits are signed.

Sounds good! I think it's better to have it changed for both `1.9` and `next-release`. 
I thought all my commits are already signed, aren't they?

yurishkuro:
we just merged 1.10

the DCO check shows some commits unsigned. You can squash them and resign the one commit.

d1mnewz:
Agree. I've just changed my signature to correct one.

content/docs/1.9/getting-started.md
```diff
@@ -119,8 +119,7 @@ Then navigate to `http://localhost:8080`.
 
 ## Migrating from Zipkin
 
-Collector service exposes Zipkin compatible REST API `/api/v1/spans` and `/api/v2/spans` for both
```
yurishkuro: why 1.9 and not 1.10 ?

d1mnewz: Actually, this is not changed much. Just made it one line instead of two lines. The content is the same.

yurishkuro: I am saying 1.9 is already N-1 release, we have 1.1o now: https://www.jaegertracing.io/docs/1.10/getting-started/#migrating-from-zipkin

d1mnewz: I see. I've changed my updates to alter v1.10 docs.

PR #241, opened 2019-04-19
Call out that Kafka is a buffer, not a storage

### LINKED ISSUE #207: Clarify Kafka Documentation, specifically that it can't be used as a lone storage option.

We were looking to quickly deploy jaeger using our existing kafka cluster as a backend. 

When deploying jaeger-query we receive an error saying kafka was write only. The Documentation for version 1.9 doesn't indicate that kafka cannot serve as the span storage for jaeger-query. The Documentation for 1.6 does point out the limited capability 

> In version 1.6.0, the Kafka storage backend implementation only supports writing data, this means you will need to use another one (with the multiple storage types feature) to be able to view the traces in the Query component.
> 
> Starting from version 1.7.0, a new component (Ingester) will be added to support reading from Kafka and storing it in another storage backend (Elasticsearch or Cassandra).

Amend the 1.9 and future documentation to note that Kafka is write-only storage and must be used in conjunction with ingester and another storage in order to support query.  It may make sense to rethink how Kafka's role is described. It's being used to queue transactions for storage, not actually as a storage backend.

yurishkuro:
Resolves #207


content/docs/1.10/deployment.md
```diff
@@ -133,7 +133,7 @@ Port  | Protocol | Function
 
 Collectors require a persistent storage backend. Cassandra and Elasticsearch are the primary supported storage backends. Additional backends are [discussed here](https://github.com/jaegertracing/jaeger/issues/638).
 
-The storage type can be passed via `SPAN_STORAGE_TYPE` environment variable. Valid values are `cassandra`, `elasticsearch`, `kafka` and `memory` (only for all-in-one binary).
+The storage type can be passed via `SPAN_STORAGE_TYPE` environment variable. Valid values are `cassandra`, `elasticsearch`, `kafka` (only as a sink) and `memory` (only for all-in-one binary).
```
epkugelmass: I'm not sure it's immediately clear what a `sink` means here. "Cannot power jaeger-query" is more explicit that kafka cannot be used (solo).

pavolloffay: I would also prefer a different name than sink, maybe a `buffer`?

PR #247, opened 2019-05-06
Replacing Architecture diagram with a svg

ledor473:
PNG are hard to edit so I re-did the Architecture using LucidChart and exported it as an SVG.

Why? I wanted to document the Ingester with Kafka in the architecture diagram, but for my upcoming talk I also wanted to modify it for a slide (to add extra information).
  - Loosely relates to https://github.com/jaegertracing/documentation/pull/241

I could not replicate and had to modify only the following 2 shapes:
- `Jaeger UI` box with rounded top corners
- `Adaptive Sampling` shape

Also, I've updated the Go logo and added C#.

I tried to make everything else pixel-perfect (including text font, color, size).

## Open Question
- I can also export it as a Visio `.vdx` to allow for easier modification over time, should I commit it as well?


yurishkuro:
could we store Lucid source files, for future updates?

ledor473:
@yurishkuro It doesn't seem like LucidChart has an export format so I tried to export it as `vdx` (Visio). It exports correctly and I'm able to re-import it without loss.
The problems begin when you have a free account, the import works but LucidChart doesn't allow to modify an imported document.
The workaround I've found is that I can share a read-only link of the graph and people can make a copy of it and modify it afterward.

I've also tried importing it inside draw.io and it wasn't great.
I'm not sure if Visio would import it successfully but I don't have it so I.

yurishkuro:
In this case I would prefer to do it  in Google Slides, since everyone has access to that and we can export individual slides to SVG.

yurishkuro:
(it's where the original diagrams came from as well)

ledor473:
[Somehow](https://desk.draw.io/support/solutions/articles/16000063693-how-to-import-lucidchart-diagrams-into-draw-io) I get a better result by copy-pasting from LucidChart to Draw.io. It's not perfect but would likely take less time to fix than re-do in Google Slides.

Would that work @yurishkuro ?

yurishkuro:
In what format does draw.io save them?

ledor473:
When I used it in the past, I was able to save the diagram on Google Drive. I think it eventually end up being some [XML format](https://support.draw.io/display/DO/Exporting+Files)

ledor473:
Also, should we merge `jaeger-ui` and `jaeger-query` as they are deployed together?

yurishkuro:
Let's use the Google slides as the source. I copied various diagrams into the deck below, from various conference talks.

https://docs.google.com/presentation/d/1JuurkQn03z0BbOEAViJBEE_WWMj6JQUML-uJm7zizvI/

pavolloffay:
Can we merge this?

yurishkuro:
I prefer we use svgs from slides before merging.

ledor473:
I should have some time next week to re-do the diagram in Google Slides 👍 not sure if it will be exportable in svg, but I'll check that as well

yurishkuro:
@ledor473 the diagram are already done in the slides, no need to redo (unless you want to edit them a bit).

ledor473:
Right, I meant I'll redo the changes in this one https://github.com/jaegertracing/documentation/pull/248 in Google Slides 👍

PR #284, opened 2019-07-25
Fix title capitalization

yurishkuro:
This makes the page titles consistently capitalized.

cc @jpkrohling

### DIFF
--- a/content/docs/1.13/client-libraries.md
+++ b/content/docs/1.13/client-libraries.md
@@ -1,8 +1,8 @@
 ---
-title: Client libraries
+title: Client Libraries
 weight: 5
 children:
-- title: Client features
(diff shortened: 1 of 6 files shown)


content/docs/1.13/client-libraries.md
```diff
@@ -1,8 +1,8 @@
 ---
-title: Client libraries
+title: Client Libraries
 weight: 5
 children:
-- title: Client features
+- title: Client Features
```
jpkrohling: How about just "Features"? It's under "Client Libraries" already.

yurishkuro: I tried that previously, but we have another Features page and it looked slightly confusing.

PR #311, opened 2019-10-11
Clarify that span ID of 0 is not valid

yurishkuro:
If we want to remain future-compatible with W3C


jpkrohling:
What's the context for this change? Is this based on the questions from @ledor473?

yurishkuro:
@jpkrohling context is https://github.com/jaegertracing/jaeger-client-cpp/issues/182

PR #335, opened 2019-12-10
Add FAQ about jaeger-agent

yurishkuro:
People often ask why they need to run the agent.

https://stackoverflow.com/questions/59153293/advantages-of-using-jaeger-agent


yurishkuro:
cc @objectiser

PR #338, opened 2019-12-18
Document self-provisioned Elasticsearch example CR

### LINKED ISSUE #336: Document how to use persistent storage with self provisioned ES on openshift

pavolloffay:
Resolves #336 

https://deploy-preview-338--jaegertracing.netlify.com/docs/next-release/operator/#self-provisioned

Once approved I will copy it to current and older releases.

### DIFF
--- a/content/docs/next-release/operator.md
+++ b/content/docs/next-release/operator.md
@@ -433,18 +433,18 @@ spec:
       resources: # <3>
         requests:
           cpu: 200m
-          memory: 2Gi
+          memory: 4Gi
         limits:
-          memory: 2Gi
+          memory: 4Gi
       redundancyPolicy: ZeroRedundancy # <4>
 ```
 <1> Number of Elasticsearch nodes. For high availability use at least 3 nodes. Do not use 2 nodes as "split brain" problem can happen.
 
 <2> Persistent storage configuration. In this case AWS `gp2` with `5Gi` size. When omitted `emptyDir` is used. Elasticsearch operator provisions `PersistentVolumeClaim` and `PersistentVolume` which are not removed with Jaeger instance. The same volumes can be mounted if Jaeger with the same name and namespace is crated. Some storages might fail in `default` namespace because of OpenShift SCC policy.
 
-<3> Resources for Elasticsearch nodes. In this case `2Gi` which is not suitable for production ready cluster. Refer to Elasticsearch documentation for memory recommendations.
+<3> Resources for Elasticsearch nodes. In this case `4Gi` which results to by default required `2Gi` of heap space. Refer to Elasticsearch [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/heap-size.html) for memory recommendations.
 
-<4> Data replication policy defines how Elasticsearch shards are replicated across data nodes in the cluster. If not specified Jaeger operator automatically determines the most appropriate replication based on number of nodes.
+<4> Data replication policy defines how Elasticsearch shards are replicated across data nodes in the cluster. If not specified Jaeger Operator automatically determines the most appropriate replication based on number of nodes.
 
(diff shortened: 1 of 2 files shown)


content/docs/next-release/operator.md
```diff
@@ -405,14 +405,51 @@ spec:
...
+```yaml
+apiVersion: jaegertracing.io/v1
+kind: Jaeger
+metadata:
+  name: simple-prod
+spec:
+  strategy: production
+  storage:
+    type: elasticsearch
+    elasticsearch:
```
pavolloffay: @jcantrill FYI

content/docs/next-release/operator.md
```diff
@@ -405,14 +405,51 @@ spec:
...
+        requests:
+          cpu: 200m
+          memory: 2Gi
+        limits:
+          memory: 2Gi
+      redundancyPolicy: ZeroRedundancy # <4>
+```
+<1> Number of Elasticsearch nodes. For high availability use at least 3 nodes. Do not use 2 nodes as "split brain" problem can happen.
+
+<2> Persistent storage configuration. In this case AWS `gp2` with `5Gi` size. When omitted `emptyDir` is used. Elasticsearch operator provisions `PersistentVolumeClaim` and `PersistentVolume` which are not removed with Jaeger instance. The same volumes can be mounted if Jaeger with the same name and namespace is crated. Some storages might fail in `default` namespace because of OpenShift SCC policy.
```
jpkrohling: Second time in a few weeks that we get bitten by this... `default` should really be just like any other namespace :/

content/docs/next-release/operator.md
```diff
@@ -405,14 +405,51 @@ spec:
...
+        size: 5Gi
+      resources: # <3>
+        requests:
+          cpu: 200m
+          memory: 2Gi
+        limits:
+          memory: 2Gi
+      redundancyPolicy: ZeroRedundancy # <4>
+```
+<1> Number of Elasticsearch nodes. For high availability use at least 3 nodes. Do not use 2 nodes as "split brain" problem can happen.
```
jpkrohling: Is there a doc page in ES documentation that talks about this problem? If so, this part could link there.

content/docs/next-release/operator.md
```diff
@@ -405,14 +405,51 @@ spec:
...
+
+<2> Persistent storage configuration. In this case AWS `gp2` with `5Gi` size. When omitted `emptyDir` is used. Elasticsearch operator provisions `PersistentVolumeClaim` and `PersistentVolume` which are not removed with Jaeger instance. The same volumes can be mounted if Jaeger with the same name and namespace is crated. Some storages might fail in `default` namespace because of OpenShift SCC policy.
+
+<3> Resources for Elasticsearch nodes. In this case `2Gi` which is not suitable for production ready cluster. Refer to Elasticsearch documentation for memory recommendations.
```
jpkrohling: What's the minimum they recommend? The name of this example is `simple-prod`, we could really give readers an idea of what's the base line for a production deployment.

content/docs/next-release/operator.md
```diff
@@ -405,14 +405,51 @@ spec:
...
+
+<4> Data replication policy defines how Elasticsearch shards are replicated across data nodes in the cluster. If not specified Jaeger operator automatically determines the most appropriate replication based on number of nodes.
```
jpkrohling: `s/operator/Operator`

content/docs/next-release/operator.md
pavolloffay: I coud not find any nice official page. It's easily searchable and there are a lot of blog posts.

content/docs/next-release/operator.md
pavolloffay: They recommend at least 2gb of heap - which means we have to set 4gb here https://www.elastic.co/guide/en/elasticsearch/reference/5.6/heap-size.html. I will change and add link

pavolloffay:
@JStickler the same docs should be probably added to prod version. We got asked how to configure ES several times.

PR #354, opened 2020-01-27
Added port forword for grpc port

mschneider82:
It was missing in the docker run statement but mentioned below in the exposed ports.


mschneider82:
should i add it to the current one?

yurishkuro:
Thank you!

PR #357, opened 2020-02-06
Updated the data format/transport table for clients

jpkrohling:
- The doc page with the data format/transport seems to be outdated. I went over the listed clients and tried to sync the data based on different types of evidence (code, test, example or documentation), resulting in this PR.

I also checked which Thrift encoding formats are supported by each client, but we don't seem to be recording this anywhere. Is this something we'd want to record somewhere?

Client | Compact | Binary
-- | -- | --
Java | ✔️ | ✔️
Go | ✔️ | ❌
C# | ✔️ | ❌
C++ | ✔️ | ❌
Python | ✔️ | ❌
Node | ❌ | ✔️

### DIFF
--- a/data/clients.yaml
+++ b/data/clients.yaml
@@ -5,8 +5,6 @@ featureGroups:
       description: Report [`jaeger.thrift`](https://github.com/jaegertracing/jaeger-idl/blob/master/thrift/jaeger.thrift), [Compact](https://github.com/apache/thrift/blob/v0.13.0/doc/specs/thrift-compact-protocol.md) protocol, over UDP
     - name: report_jaeger_thrift_binary_udp
       description: Report [`jaeger.thrift`](https://github.com/jaegertracing/jaeger-idl/blob/master/thrift/jaeger.thrift), [Binary](https://github.com/apache/thrift/blob/v0.13.0/doc/specs/thrift-binary-protocol.md) protocol, over UDP
-    - name: report_jaeger_thrift_compact_http
-      description: Report [`jaeger.thrift`](https://github.com/jaegertracing/jaeger-idl/blob/master/thrift/jaeger.thrift), [Compact](https://github.com/apache/thrift/blob/v0.13.0/doc/specs/thrift-compact-protocol.md) protocol, over TCP (HTTP)
     - name: report_jaeger_thrift_binary_http
       description: Report [`jaeger.thrift`](https://github.com/jaegertracing/jaeger-idl/blob/master/thrift/jaeger.thrift), [Binary](https://github.com/apache/thrift/blob/v0.13.0/doc/specs/thrift-binary-protocol.md) protocol, over TCP (HTTP)
     - name: report_zipkin_thrift_binary_http
@@ -141,7 +139,6 @@ clients:
     features:
       report_jaeger_thrift_compact_udp: yes
       report_jaeger_thrift_binary_udp: no
-      report_jaeger_thrift_compact_http: no
       report_jaeger_thrift_binary_http: yes
       report_zipkin_thrift_binary_http: yes
       propagation_uber: yes
@@ -184,7 +181,6 @@ clients:
     features:
       report_jaeger_thrift_compact_udp: yes
       report_jaeger_thrift_binary_udp: no
-      report_jaeger_thrift_compact_http: no
       report_jaeger_thrift_binary_http: yes
       report_zipkin_thrift_binary_http: yes
       propagation_uber: yes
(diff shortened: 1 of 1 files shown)


jpkrohling:
> It's news to me that Java client can do both - are you sure?

I haven't tested myself, but looks like we do offer this option for people who want to customize the tracer: 

https://github.com/jaegertracing/jaeger-client-java/blob/74adce4c4b4c507ccfb010ffb06c95973cc97c6a/jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/senders/ThriftSenderBase.java#L48-L69

I'll change the PR to include the protocol type in a separate table

data/clients.yaml
```diff
@@ -178,7 +187,9 @@ clients:
     features:
       report_jaeger_thrift_udp: yes
       report_jaeger_thrift_http: yes
-      report_zipkin_thrift_http: no
+      report_zipkin_thrift_http: yes # https://github.com/jaegertracing/jaeger-client-java/tree/v0.35.5/jaeger-zipkin#thrift
+      report_thrift_compact: yes
+      report_thrift_binary: yes
```
yurishkuro: I think this is confusing in this form, because it's not clear how binary/compact protocol is related to udp/http transport. I believe HTTP always uses `binary` on the backend, and Java client is hardcoded to use `compact` over UDP: https://github.com/jaegertracing/jaeger-client-java/blob/74adce4c4b4c507ccfb010ffb06c95973cc97c6a/jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/senders/UdpSender.java#L47

yurishkuro: Maybe the matrix should be explicit (even though it gets somewhat verbose):
```
      report_jaeger_thrift_compact_udp: yes
      report_jaeger_thrift_binary_udp: no ### yes for Node only
      report_jaeger_thrift_compact_http: #### always no for all clients
      report_jaeger_thrift_binary_http: yes
      report_zipkin_thrift_binary_http:
```

jpkrohling: > I believe HTTP always uses binary on the backend

This seems indeed to be the case:
https://github.com/jaegertracing/jaeger/blob/29e71310f3f275def3e07d257901b6e89ad0818e/cmd/collector/app/http_handler.go#L82

https://github.com/apache/thrift/blob/ded326101af3c6c9daad9814ce6404d385f36a1f/lib/go/thrift/deserializer.go#L31-L39

jpkrohling: Does it still make sense to have network protocol + thrift format, given that all HTTP clients are sending Thrift Binary?

yurishkuro: yes, because UDP uses both binary and compact, so I would rather be consistent and have report_jaeger_thrift_binary_http, even though we won't have any report_jaeger_thrift_compact_http

jpkrohling:
PR updated. For reference, here's how the table looked like before the last commit:

image

and this is how it looks now:

image

data/clients.yaml
```diff
@@ -1,12 +1,16 @@
...
-      description: Report [`jaeger.thrift`](https://github.com/jaegertracing/jaeger-idl/blob/master/thrift/jaeger.thrift) over UDP
-    - name: report_jaeger_thrift_http
-      description: Report [`jaeger.thrift`](https://github.com/jaegertracing/jaeger-idl/blob/master/thrift/jaeger.thrift) over HTTP
-    - name: report_zipkin_thrift_http
-      description: Report Zipkin Thrift over HTTP
+    - name: report_jaeger_thrift_compact_udp
+      description: Report [`jaeger.thrift`](https://github.com/jaegertracing/jaeger-idl/blob/master/thrift/jaeger.thrift), [Compact](https://github.com/apache/thrift/blob/v0.13.0/doc/specs/thrift-compact-protocol.md) protocol, over UDP
+    - name: report_jaeger_thrift_binary_udp
+      description: Report [`jaeger.thrift`](https://github.com/jaegertracing/jaeger-idl/blob/master/thrift/jaeger.thrift), [Binary](https://github.com/apache/thrift/blob/v0.13.0/doc/specs/thrift-binary-protocol.md) protocol, over UDP
+    - name: report_jaeger_thrift_compact_http
```
yurishkuro: do we need this line? no client uses this format, and I am not even sure if the server supports it

jpkrohling: I actually confirmed in a previous comment that the server does not support Thrift Compact via HTTP, but you said that it's better if the matrix is explicit. If you now prefer to not have it listed at all, I can remove it, especially as it's also my preference ;-)

yurishkuro: let's remove what doesn't exist.

jpkrohling: Done!

PR #387, opened 2020-03-13
Fix value type in Jaeger Operator default sampling example

### LINKED ISSUE #386: Incorrect value type used in kubernetes operator sampler param example

Configure a default sampling configuration using Jaeger operator

The documentation (section is [Defining Sampling Strategies](https://www.jaegertracing.io/docs/1.16/operator/#defining-sampling-strategies)) shows an example of setting a 50% sampling rate as an integer (50).  This should be a double (0.5).

Change 50 to 0.5.  I will send a PR as it's a small change.

linjmeyer:
Resolves #386 

- Jaeger Operator default sampling configuration uses integer value type instead of double

- Value of 50 was changed to .5

### DIFF
--- a/content/docs/next-release/operator.md
+++ b/content/docs/next-release/operator.md
@@ -859,7 +859,7 @@ spec:
     options:
       default_strategy:
         type: probabilistic
-        param: 50
+        param: .5
 ```
 
 This example defines a default sampling strategy that is probabilistic, with a 50% chance of the trace instances being sampled.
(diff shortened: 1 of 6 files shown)


content/docs/next-release/operator.md
```diff
@@ -859,7 +859,7 @@ spec:
     options:
       default_strategy:
         type: probabilistic
-        param: 50
+        param: .5
```
yurishkuro, line 862: I am not able to parse this form:
```
failed to unmarshal strategies: invalid character '.' looking for beginning of value
```

Please update to `0.5`

linjmeyer: Done!

PR #418, opened 2020-07-04
Added note about jaeger_agent_client_stats_spans_dropped_total

joe-elliott:
##  Which problem is this PR solving?
- Added a reference to a useful metric for performance tuning.

- While doing research on dropped spans I found `jaeger_agent_client_stats_spans_dropped_total` and felt like it should be  included in this doc.

Should I PR this change against all past versions of this doc?


yurishkuro:
thank!

PR #429, opened 2020-08-23
Fix --es.server-urls flag example

yyyogev:
This flag is wrong and causes an error when try to use it.

documentation typo

## Short description of the changed
 --es.server.urls  ->  --es.server-urls


yurishkuro:
I backported the fix to 1.18 docs as well.

PR #443, opened 2020-10-01
Add missing redirects

yurishkuro:
We've added quite a few /docs/ pages, but did not update the redirects file so that `/docs/{name}` URLs work by redirecting to the latest page.

The redirects can be generated with this command:
```shell
ls content/docs/1.19 | grep -v _index | sed 's|^\(.*\).md$|/docs/\1     /docs/{{ $latest }}/\1|g'
```


jpkrohling (approved):
I had no idea we have those in place :-)

PR #446, opened 2020-10-02
Use shortcodes for red/green check marks

yurishkuro:
GitHub-style checkboxes don't translate that well. This introduces shortcodes that render emojis that we already used in the client feature matrix.


jpkrohling:
Should these here be changed to use the new macro as well?

https://github.com/jaegertracing/documentation/blob/af4c87f6f163eabe06bd9c62108c2d6daeed8554/themes/jaeger-docs/layouts/shortcodes/featuresTable.html#L22-L25

yurishkuro:
I don't think it works like that, the shortcodes can be used in markdown files, they cannot be used in the html partials.

jpkrohling (approved):

PR #454, opened 2020-10-21
Update getting-started.md

ronaldpetty:
I did not confirm on all hosts, only using Mac.  "make install" does not work out of box, it also diverges from the actual hot rod page (where make install is removed).

https://github.com/jaegertracing/jaeger/tree/master/examples/hotrod

Unclear where to place this (1.20, latest, etc.).

### DIFF
--- a/content/docs/1.20/getting-started.md
+++ b/content/docs/1.20/getting-started.md
@@ -94,7 +94,6 @@ mkdir -p $GOPATH/src/github.com/jaegertracing
 cd $GOPATH/src/github.com/jaegertracing
 git clone git@github.com:jaegertracing/jaeger.git jaeger
 cd jaeger
-make install
 go run ./examples/hotrod/main.go all
 ```
 #### From docker


ronaldpetty:
I'll leave PR here and someone else can add this to their work, too much effort (DCO) for potential typo.  Maybe one day GitHub will make it work directly.

content/docs/1.20/getting-started.md
```diff
@@ -94,7 +94,6 @@ mkdir -p $GOPATH/src/github.com/jaegertracing
 cd $GOPATH/src/github.com/jaegertracing
 git clone git@github.com:jaegertracing/jaeger.git jaeger
 cd jaeger
-make install
```
yurishkuro, line 97: We can also remove the two commands with GOPATH, since we switched to go modules and the repo can reside in any location.

ronaldpetty:
Similar updates might be good on blog as well.

https://medium.com/opentracing/take-opentracing-for-a-hotrod-ride-f6e3141f7941

PR #455, opened 2020-10-27
Add OTLP receiver port to OpenTelemetry page

bboreham:
This makes it easier for users to configure their OpenTelemetry clients.

Mention that the OTLP receiver is on by default, and which port it listens on.
My understanding of the previous text was that I would have to configure it myself.

I found its configuration in the source code here: https://github.com/jaegertracing/jaeger/blob/3c0505fc779e2d054c23304134bef4f22ce9ce75/cmd/opentelemetry/app/defaultconfig/default_config.go#L163

and here: https://github.com/open-telemetry/opentelemetry-collector/blob/7dd853ab95834619169360fa2abbb981af42f061/receiver/otlpreceiver/factory.go#L62

### DIFF
--- a/content/docs/next-release/opentelemetry.md
+++ b/content/docs/next-release/opentelemetry.md
@@ -26,6 +26,7 @@ The Jaeger OpenTelemetry binaries are **almost** backward compatible with the cu
 
 The differences are:
 
+* OTLP receiver listening on port 55680
 * Health check port changed to `13133`
 * Not all current Jaeger flags are exposed (e.g. health check port)
 * Exposed metrics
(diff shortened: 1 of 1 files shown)


joe-elliott (commented):
It is true that the agent and collector currently start an OTLP receiver, but I don't know if that's intentional or not. 

We are currently working on the details of what services the agent/collector should start by default.  It is my opinion that the OTLP receiver should not be started by default, but instead should require a flag.

yurishkuro:
@joe-elliott what is your concern with starting OTLP receiver by default? That seems consistent with the approach that Jaeger components are extensions of OTEL default components, meaning they can be used as a drop-in replacement with some extra functionality.

content/docs/next-release/opentelemetry.md
```diff
@@ -26,6 +26,7 @@ The Jaeger OpenTelemetry binaries are **almost** backward compatible with the cu
 
 The differences are:
 
+* OTLP receiver listening on port 55680
```
yurishkuro, line 29: Is this the default port in OTEL?  Not a great choice picking a port from ephemeral range, could run into random port conflicts.

joe-elliott: Yup, 55680 is default.

https://github.com/open-telemetry/opentelemetry-collector/tree/master/receiver/otlpreceiver

I had not thought of that before, but you are correct that could conflict with dynamic ports.

joe-elliott:
I suppose my only real concern was that adding the otlp receiver to both the agent/collector by default would prevent them from starting on cleanly in the same network namespace.

I also did not realize we were ok with extending the features of the agent/collector by default, but am generally fine with it.

yurishkuro (approved):

content/docs/next-release/opentelemetry.md
bboreham: I filed https://github.com/open-telemetry/opentelemetry-specification/issues/1148.
Note that Jaeger is using at least one port (6831) assigned to someone else.

PR #458, opened 2020-11-17
Remove the other option that will break on unbound variables

joe-elliott:
- Related to #457 
- Removed the other option that will cause a script to fail if there are unbound variables.


yurishkuro (approved):

yurishkuro:
@joe-elliott please use a fork for PRs, we recommend this even for maintainers, instead of creating public branches in the main repo.

PR #462, opened 2020-12-02
Clarify baggage value encoding

yurishkuro:
The old text was incorrectly describing baggage values as always URL-encoded.

### DIFF
--- a/content/docs/1.21/client-libraries.md
+++ b/content/docs/1.21/client-libraries.md
@@ -138,7 +138,7 @@ uberctx-key2: value2
 
 #### Value Encoding
 
-OpenTracing defines two formats for plan text headers: `HTTP_HEADERS` and `TEXT_MAP`. The former was introduced to deal with restrictions imposed by the HTTP protocol on the context of the headers, whereas the latter does not impose any restrictions, e.g. it can be used with Kafka Record Headers. The main difference between these two formats in the Jaeger SDKs is that the baggage values are URL-encoded when using the `HTTP_HEADERS` propagation format.
+OpenTracing defines two formats for plain text headers: `HTTP_HEADERS` and `TEXT_MAP`. The former was introduced to deal with restrictions imposed by the HTTP protocol on the context of the headers, whereas the latter does not impose any restrictions, e.g. it can be used with Kafka Record Headers. The main difference between these two formats in the Jaeger SDKs is that the baggage values are URL-encoded when using the `HTTP_HEADERS` propagation format.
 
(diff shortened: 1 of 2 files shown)


content/docs/1.21/client-libraries.md
```diff
@@ -131,6 +136,22 @@ uberctx-key1: value1
 uberctx-key2: value2
 ```
 
+#### Value Encoding
+
+OpenTracing defines two formats for plan text headers: `HTTP_HEADERS` and `TEXT_MAP`. The former was introduced to deal with restrictions imposed by the HTTP protocol on the context of the headers, whereas the latter does not impose any restrictions, e.g. it can be used with Kafka Record Headers. The main difference between these two formats in the Jaeger SDKs is that the baggage values are URL-encoded when using the `HTTP_HEADERS` propagation format.
```
jpkrohling, line 141:
```suggestion
OpenTracing defines two formats for plain text headers: `HTTP_HEADERS` and `TEXT_MAP`. The former was introduced to deal with restrictions imposed by the HTTP protocol on the context of the headers, whereas the latter does not impose any restrictions, e.g. it can be used with Kafka Record Headers. The main difference between these two formats in the Jaeger SDKs is that the baggage values are URL-encoded when using the `HTTP_HEADERS` propagation format.
```

content/docs/next-release/client-libraries.md
```diff
@@ -131,6 +136,22 @@ uberctx-key1: value1
 uberctx-key2: value2
 ```
 
+#### Value Encoding
+
+OpenTracing defines two formats for plan text headers: `HTTP_HEADERS` and `TEXT_MAP`. The former was introduced to deal with restrictions imposed by the HTTP protocol on the context of the headers, whereas the latter does not impose any restrictions, e.g. it can be used with Kafka Record Headers. The main difference between these two formats in the Jaeger SDKs is that the baggage values are URL-encoded when using the `HTTP_HEADERS` propagation format.
```
jpkrohling, line 141:
```suggestion
OpenTracing defines two formats for plain text headers: `HTTP_HEADERS` and `TEXT_MAP`. The former was introduced to deal with restrictions imposed by the HTTP protocol on the context of the headers, whereas the latter does not impose any restrictions, e.g. it can be used with Kafka Record Headers. The main difference between these two formats in the Jaeger SDKs is that the baggage values are URL-encoded when using the `HTTP_HEADERS` propagation format.
```

jpkrohling (commented):
Once `plan text` is fixed, it's ready to be merged.

PR #440, opened 2020-09-28
Fix broken links

celestehorgan:
Signed-off-by: Celeste Horgan <celeste@cncf.io>

Fixes links mentioned in #437.

PR #466, opened 2020-12-15
Use latest docs version for canonical URL

jaylinski:

With this change, the latest docs version is used as canonical URL to improve indexing in search engines.

The canonical URL should point to the latest
version of the docs, so Google will index the
most current documentation.

Previous related change: https://github.com/jaegertracing/documentation/pull/440

https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls#rel-canonical-link-method

### Motivation

I copied parts of this canonical-logic for another docs-site and noticed this issue.

### DIFF
--- a/themes/jaeger-docs/layouts/partials/meta.html
+++ b/themes/jaeger-docs/layouts/partials/meta.html
@@ -14,15 +14,17 @@
 
 <!-- OpenGraph metadata -->
 <meta property="og:title" content="{{ $title }}" />
+<meta property="og:url" content="{{ .Permalink }}" />
 {{ if $isDoc }}
+<meta property="og:type" content="documentation" />
 {{ $version := index (split .Path "/") 1 }}
 {{ $latest := site.Params.latest }}
-{{ $url := replace .RelPermalink $version "latest" }}
-<meta property="og:type" content="documentation" />
-<meta property="og:url" content="{{ .Permalink }}" />
-<link rel="canonical" href="{{ .Permalink }}" />
+{{ $url := .RelPermalink }}
+{{ if in site.Params.versions $version }}
+{{ $url = replace .RelPermalink $version $latest }}
+{{ end }}
+<link rel="canonical" href="{{ $url }}" />
 {{ else }}
-<meta property="og:url" content="{{ .Permalink }}" />
 <link rel="canonical" href="{{ .Permalink }}" />
 {{ end }}
 <meta property="og:locale" content="en_US" />


themes/jaeger-docs/layouts/partials/meta.html
```diff
@@ -14,15 +14,17 @@
...
 {{ if $isDoc }}
+<meta property="og:type" content="documentation" />
 {{ $version := index (split .Path "/") 1 }}
 {{ $latest := site.Params.latest }}
-{{ $url := replace .RelPermalink $version "latest" }}
-<meta property="og:type" content="documentation" />
-<meta property="og:url" content="{{ .Permalink }}" />
-<link rel="canonical" href="{{ .Permalink }}" />
+{{ $url := .RelPermalink }}
+{{ if in site.Params.versions $version }}
```
jaylinski, line 23: This check prevents the issue described in #440.

yurishkuro:
Are you sure this is working? In the preview I am seeing link with numbered version:

```
<link rel="canonical" href="https://deploy-preview-466--jaegertracing.netlify.app/docs/1.21/deployment/">
```

jaylinski:
> Are you sure this is working? In the preview I am seeing link with numbered version:
> 
> ```
> <link rel="canonical" href="https://deploy-preview-466--jaegertracing.netlify.app/docs/1.21/deployment/">
> ```

That's expected, since `1.21` is the current latest version. All canonical links will change as soon as a new version is released.

yurishkuro:
Maybe I do not understand the intention of this change then, I thought canonical would point to "/latest/", which is stable.

There's another argument that fooling search engine to point to a different page is not ideal to begin with, e.g. I search for a certain term that only existed in version N, Google shows this term in the results, but when I visit the page it cannot be found there.

jaylinski:
@yurishkuro I can change it back to "latest" if you prefer this. The "latest"-version currently redirects to "1.21" and I thought it would be better to use the effective URL.

About fooling search engines: according to the documentation (https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls) Google only uses this information to prioritize its crawling.

I'm certain that the current usage of the canonical-link doesn't make sense, but I'm not sure what the best solution is. I searched for other Google products with a similar versioned documentation in order to copy their canonical-logic, but couldn't find an example.

yurishkuro:
If this only drives crawling prioritization, then it seems `/latest/` would be always preferred in priority to `/1.21/` - if we do a new release, then crawling of `/latest/` will be immediately accessing the latest docs, whereas crawling `/1.21` will be indexing stale docs AND discovering that the canonical is now different.

I agree that the current canonical doesn't make sense, since it just points to the exact URL of the page, including old versions.

yurishkuro:
Anyway, I don't have a strong opinion - requesting `/latest/` will result in a redirect anyway, don't know what it means for crawlers. I'm fine to go with your change.

yurishkuro (approved):

PR #473, opened 2021-02-02
Add cloudnative101.dev tutorials

yurishkuro:
These seem pretty good and focused on Jaeger.


jpkrohling (approved):
Nice find!

jpkrohling:
What's this release check? It seems to be stuck?

yurishkuro:
I removed the release check from required since it only runs for specific tags.

PR #488, opened 2021-03-24
Fixed a dead link in the operator docs

joe-elliott:
- There was a dead link to an example yaml in the docs. This updates the link.


yurishkuro (approved):
we never finished the dead link detector in the CI :-(

PR #492, opened 2021-04-08
fix: Replace deprecated (and removed) agent tags option

julienvey:
jaeger.tags option has been deprecated in 1.16 (see #1854) and
removed definitively in 1.22 (See #2753)

Signed-off-by: Julien Vey <vey.julien@gmail.com>


yurishkuro (approved):
thanks

PR #494, opened 2021-04-14
fix: Use valid yaml comments in examples

julienvey:

`//` is not valid comment in yaml, replace with `#` to make it easy to copy paste


albertteoh (approved):

albertteoh:
Thanks!

PR #500, opened 2021-04-27
Add esc command not found Troubleshooting guide

albertteoh:
Signed-off-by: albertteoh <albert.teoh@logz.io>

- Address the frequent number of similar questions surrounding failing builds due to `esc: Command not found` such as https://github.com/jaegertracing/jaeger/issues/2963.

- Add instructions to address this problem in Troubleshooting guide.
- I understand there was an attempt to use the built-in "embed" package but not sure of its status right now. If we're still planning to go ahead with using "embed" I'm happy to abandon this PR.

### DIFF
--- a/content/docs/next-release/troubleshooting.md
+++ b/content/docs/next-release/troubleshooting.md
@@ -105,3 +105,18 @@ For Visual Studio Code, you need to have the following configuration at the root
             }
         ]
     }
+
+## make: esc: Command not found
+
+Make sure `esc` is installed:
+```
+make install-tools
+```
+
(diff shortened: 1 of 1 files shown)


yurishkuro:
I think a better user experience is to change Makefile to check for `esc` and print a similar warning if not found than expect them to find this in the docs. The printed warning can either contain all of the hints or just point to a Discussion link that explains.

albertteoh:
Yup, makes a lot of sense to avoid the extra few hops to find this troubleshooting doc. Thanks!

PR #502, opened 2021-05-21
Update required Go-Version from 1.11 to 1.16.

cyber-moon:
- The documentation mentions a too old Go-Version as prerequisite for the HotROD-Installation. The latest HotROD-Version uses the 'embed'-package from go, which is only available in Go-Version 1.16 or higher.

- Update required Go-Version from 1.11 to 1.16.


yurishkuro (approved):
Thanks. I modified slightly to not mention exact Go version directly.

yurishkuro:
oops, your commit wasn't signed, I will open another PR.

PR #511, opened 2021-07-07
Fix release by allowing force push on protected branch

### LINKED ISSUE #505: Automated release failing
The automated release failed and seems to have been failing for quite some time, apparently because the script can't push directly to master (correctly), and because changes require one review before being merged (also correctly). Therefore, I generated the docs locally. We need to figure out what to do in the future: make the release script generate the pull request vs. push directly to master.

_Originally posted by @jpkrohling in https://github.com/jaegertracing/documentation/issues/504#issuecomment-855695157_

pavolloffay:
Resolves #505 

There are two ways to solve this:
1. remove master from protected branches
2. allow force push to protected master branch

This PR chooses 2.

https://github.com/jaegertracing/documentation/settings/branch_protection_rules/1620218


jpkrohling:
Can't we make it open a PR instead? This way, we don't need to allow force-push to the main branch, which is dangerous.

pavolloffay:
We can but that requires more changes...

PR #526, opened 2021-09-29
Update for Adaptive Sampling

joe-elliott:
- Adds support SAMPLING_TYPE in cli doc generation and generates collector docs for sampling-type adaptive and static.
- Updates details on the sampling page about using adaptive sampling


yurishkuro:
do you expect something here?
image

joe-elliott:
Yes. To get this to work I pulled `jaegertracing/jaeger-*:latest` images and tagged them locally as `1.27`. Then I ran the release script:

```
DRY_RUN=true TRAVIS_TAG=release-1.27.0 ./travis/release.sh
```

to simulate a 1.27 release. The script then used the latest docker images as if they were 1.27 which includes support for the adaptive sampling type env var. This worked for me locally.  Can you share details about how you got the empty sections?

yurishkuro:
> This worked for me locally. Can you share details about how you got the empty sections?

oh, I just looked at `next-version` in the netlify preview. Makes sense that it wouldn't have the data, it's a bit tricky to coordinate.

content/docs/next-release/sampling.md
```diff
@@ -21,7 +21,7 @@ Adaptive sampler is a composite sampler that combines two functions:
...
   * It supports a minimum guaranteed rate of sampling, such as always allowing up to N {{< tip "traces" "trace" >}} per seconds and then sampling anything above that with a certain probability (everything is per-operation, not per-service).
 
-Per-operation parameters can be configured statically or pulled periodically from Jaeger backend with the help of Remote sampler. Adaptive sampler is designed to work with the upcoming [Adaptive Sampling](https://github.com/jaegertracing/jaeger/issues/365) feature of the Jaeger backend.
+Per-operation parameters can be configured statically or pulled periodically from Jaeger backend with the help of Remote sampler. In order to use adaptive sampling set the environment variable `SAMPLING_TYPE=adaptive` on the collector. Currently this is only available when paired with a Cassandra `SPAN_STORAGE_TYPE`. If you're just getting started please review `sampling.initial-sampling-probability` and `sampling.target-samples-per-second`, and see the [cli parameters](/docs/latest/cli/#jaeger-collector-adaptive) for details on advanced configuration.
```
yurishkuro, line 24: I don't mind merging like this, but I think we need a bit more documentation explaining how the thing actually works. I would also separate the section into two parts, one (existing) describing just the per-operation functionality (which is not technically "adaptive"), and the actual new adaptive part.

joe-elliott: The description of static sampling is basically here:

https://www.jaegertracing.io/docs/1.26/sampling/#collector-sampling-configuration

but agree the way this doc is organized makes things a bit confusing. We should probably move the adaptive sampling under "collector sampling configuration" and have two sections there (one for adaptive and one for static). I will update this PR with some structural changes.

joe-elliott: changes made

content/docs/next-release/sampling.md
```diff
@@ -12,18 +12,13 @@ When using configuration object to instantiate the tracer, the type of sampling
...
-  * It supports a minimum guaranteed rate of sampling, such as always allowing up to N {{< tip "traces" "trace" >}} per seconds and then sampling anything above that with a certain probability (everything is per-operation, not per-service).
+## Collector Sampling Configuration
 
-Per-operation parameters can be configured statically or pulled periodically from Jaeger backend with the help of Remote sampler. Adaptive sampler is designed to work with the upcoming [Adaptive Sampling](https://github.com/jaegertracing/jaeger/issues/365) feature of the Jaeger backend.
+If your clients are configured to use remote sampling then sampling rates can be centrally controlled via the collectors. In a remote sampling setup a json document is served to the Jaeger client that describes endpoints and their sampling probabilities. This document can be generated two different ways: [statically](#static-sampling) or [dynamically](#adaptive-sampling). The method of document generation is controlled by the environment variable `SAMPLING_TYPE` which can be set to either `static`(default) or `adaptive`.
```
yurishkuro, line 19: > `[statically](#static-sampling)`

just thinking out loud - the name "static" is actually a misnomer. It used to be accurate but now this same file can be reloaded on changes from the file system, so "static" sounds misleading. We may want to rename the value to `file` instead of `static`, and change the docs here accordingly to not even mention `static`.

> environment variable `SAMPLING_TYPE`

I may have missed it, but this var doesn't seem to get printed in `env` command
```shell
$ SPAN_STORAGE_TYPE=memory go run ./cmd/collector env
...
```
Since we didn't do a release with it, I would strongly suggest renaming SAMPLING_TYPE -> SAMPLING_CONFIG_TYPE

content/docs/next-release/sampling.md
```diff
@@ -83,3 +78,12 @@ In the above example:
...
+  * It supports a minimum guaranteed rate of sampling, such as always allowing up to N {{< tip "traces" "trace" >}} per seconds and then sampling anything above that with a certain probability (everything is per-operation, not per-service).
+
+Adaptive sampling requires a central store for the collectors to coordinate information about spans received/second and current sampling rates. Currently this has only been implemented for Cassandra and adaptive sampling shares the same storage as your span storage. Right now adaptive sampling will only work if `SPAN_STORAGE_TYPE` is set to Cassandra and the [relevant tables](https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/cassandra/schema/v003.cql.tmpl#L206-L229) have been created. If you're just getting started please review `sampling.initial-sampling-probability` and `sampling.target-samples-per-second`, and see the [cli parameters](/docs/latest/cli/#jaeger-collector-adaptive) for details on advanced configuration.
```
yurishkuro, line 89: Suggested text:

Adaptive sampling works in the Jaeger collector by observing the spans received from services and recalculating sampling probabilities for each service/endpoint combination to ensure that the volume of collected traces matches `--sampling.target-samples-per-second`. When a new service or endpoint is detected, it is initially sampled with `--sampling.initial-sampling-probability` until enough data is collected to calculate the rate appropriate for the traffic going through the endpoint.

yurishkuro: Once we merge this I would like to write a more detailed description of the actual algorithm.

joe-elliott:
- `SAMPLING_TYPE` env var -> `SAMPLING_CONFIG_TYPE`
- Static Sampling -> File Sampling
  - Improved description with details about loading via http and `--sampling.strategies-reload-interval`
- Replaced adaptive sampling description with suggestion.

content/docs/next-release/sampling.md
```diff
@@ -12,20 +12,15 @@ When using configuration object to instantiate the tracer, the type of sampling
...
 
-## Collector Sampling Configuration
+### File Sampling
 
-Collectors can be instantiated with static sampling strategies (which are propagated to the respective service if configured with Remote sampler) via the `--sampling.strategies-file` option. This option requires a path to a json file which defines the sampling strategies.
+Collectors can be instantiated with static sampling strategies (which are propagated to the respective service if configured with Remote sampler) via the `--sampling.strategies-file` option. This option requires a path to a json file or an http url which indicates the location of the sampling strategies file. Also note that `--sampling.strategies-reload-interval` allows for periodic reloading of the indicated file.
```
yurishkuro, line 23:
```suggestion
Collectors can be instantiated with the `--sampling.strategies-file` option that points to a file containing sampling strategies to be served to Jaeger clients. The option's value can contain a path to a JSON file, which will be automatically reloaded if its contents change, or an HTTP URL from where the file will be periodically retrieved, with reload frequency controlled by the `--sampling.strategies-reload-interval` option.
```

PR #539, opened 2021-11-30
Add metrics storage support to auto-generated CLI documentation

albertteoh:
Signed-off-by: albertteoh <see.kwang.teoh@gmail.com>

- Relates to https://github.com/jaegertracing/jaeger/issues/2954

- Adds support for metrics-storage to the auto-generated documentation.
- Minor improvements:
  - Taking advantage of python 3 features such as f-strings along with some simplifications.
  - Fix variable names and titles in template.

## Testing

Generated CLI docs against version 1.28 to allow output comparison.

```
# Only prometheus-related files were added.
$ diff -q /home/go/src/github.com/albertteoh/documentation/data/cli/1.28/ /tmp/docs/data/cli/1.28
Files /home/go/src/github.com/albertteoh/documentation/data/cli/1.28/config.json and /tmp/docs/data/cli/1.28/config.json differ
Only in /tmp/docs/data/cli/1.28: jaeger-all-in-one-prometheus.yaml
Only in /tmp/docs/data/cli/1.28: jaeger-query-prometheus.yaml

# Only prometheus flags added to these files.
$ diff /home/go/src/github.com/albertteoh/documentation/data/cli/1.28/jaeger-all-in-one.yaml /tmp/docs/data/cli/1.28/jaeger-all-in-one-prometheus.yaml
150a151,177
> - name: prometheus.connect-timeout
>   default_value: 30s
>   usage: |
>     The period to wait for a connection to Prometheus when executing queries.
> - name: prometheus.server-url
>   default_value: http://localhost:9090
>   usage: |
>     The Prometheus server's URL, must include the protocol scheme e.g. http://localhost:9090
> - name: prometheus.tls.ca
>   usage: |
>     Path to a TLS CA (Certification Authority) file used to verify the remote server(s) (by default will use the system truststore)
> - name: prometheus.tls.cert
>   usage: |
>     Path to a TLS Certificate file, used to identify this process to the remote server(s)
> - name: prometheus.tls.enabled
>   default_value: "false"
>   usage: Enable TLS when talking to the remote server(s)
> - name: prometheus.tls.key
>   usage: |
>     Path to a TLS Private Key file, used to identify this process to the remote server(s)
> - name: prometheus.tls.server-name
>   usage: |
>     Override the TLS server name we expect in the certificate of the remote server(s)
> - name: prometheus.tls.skip-host-verify
>   default_value: "false"
>   usage: |
>     (insecure) Skip server's certificate chain and host name verification
```

Copied generated doc files to `documentation/data/cli/1.28/` and ran `make develop`. The following are relevant screenshots of the rendered changes made in this PR:

Screen Shot 2021-11-30 at 10 31 50 pm

Screen Shot 2021-11-30 at 10 31 28 pm

Screen Shot 2021-11-30 at 10 32 38 pm
Screen Shot 2021-11-30 at 10 32 33 pm
Screen Shot 2021-11-30 at 10 32 19 pm
Screen Shot 2021-11-30 at 10 32 15 pm
Screen Shot 2021-11-30 at 10 32 04 pm

### DIFF
--- a/scripts/gen-cli-data.py
+++ b/scripts/gen-cli-data.py
@@ -3,31 +3,41 @@ import sys
 import json
 
 if len(sys.argv) < 3:
-    print("You must supply a Jaeger version as the first argument and the output path as the second")
+    print("You must supply a Jaeger version as the first argument and "
+          "the output path as the second")
     sys.exit(1)
 
-jaeger_ver=sys.argv[1]
-output_path=sys.argv[2]
+jaeger_ver = sys.argv[1]
+output_path = sys.argv[2]
 
-with open("%s/data/cli/%s/config.json" % (output_path, jaeger_ver), 'r') as f:
-    cfg=json.load(f)
+with open(f"{output_path}/data/cli/{jaeger_ver}/config.json", 'r') as f:
+    cfg = json.load(f)
 
-def generate(tool, storage='', sampling=''):
-    print('Generating YAML for {}, storage {} and sampling {}'.format(tool, storage, sampling))
 
-    volume='%s/data/cli/%s:/data' % (output_path, jaeger_ver)
-    docker_img='all-in-one' if tool == 'jaeger-all-in-one' else tool
-    docker_image="jaegertracing/%s:%s" % (docker_img, jaeger_ver)
-    storageOrSampling=storage or sampling
+def generate(tool, **kwargs):
+    if len(kwargs) > 1:
+        print(f"Expected 1 feature, got {len(kwargs)}.")
+        sys.exit(1)
 
-    cmd=" ".join([
+    feature_name = ""
+    for _, feature_name in kwargs.items():
+        break
+
+    print(f'Generating YAML for tool: {tool}, feature: {feature_name}')
+
+    volume = f'{output_path}/data/cli/{jaeger_ver}:/data'
+    docker_img = 'all-in-one' if tool == 'jaeger-all-in-one' else tool
+    docker_image = f"jaegertracing/{docker_img}:{jaeger_ver}"
+
+    cmd = " ".join([
         "docker run",
         "--rm",
         "--interactive",
         "--privileged",
-        "--volume {}".format(volume),
-        "-e SPAN_STORAGE_TYPE={}".format(storage),
-        "-e SAMPLING_CONFIG_TYPE={}".format(sampling),
+        f"--volume {volume}",
+        f"-e SPAN_STORAGE_TYPE={kwargs.get('storage', '')}",
+        f"-e SAMPLING_CONFIG_TYPE={kwargs.get('sampling', '')}",
+        f"-e METRICS_STORAGE_TYPE={kwargs.get('metrics_storage', '')}",
         docker_image,
         "docs",
         "--format=yaml",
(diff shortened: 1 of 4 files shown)


themes/jaeger-docs/layouts/shortcodes/cli/tools-list.html
```diff
@@ -49,17 +50,32 @@ <h2 id="{{ $tool }}">
...
       <ul>
         {{ range $sampling_types }}
-          {{ $storage := . }}
-          {{ $file    := printf "%s-%s" $tool $storage }}
+          {{ $sampling := . }}
+          {{ $file    := printf "%s-%s" $tool $sampling }}
           <li>
               <a href="#{{ $file }}">
-                  {{ $tool }} with <code>{{ $storage }}</code>
+                  {{ $tool }} with <code>{{ $sampling }}</code> sampling
```
albertteoh, line 57: Before: "... with adaptive"
After: "... with adaptive sampling"

themes/jaeger-docs/layouts/shortcodes/cli/tools-list.html
```diff
@@ -80,11 +96,25 @@ <h3 id="{{ $file }}">
 
       <!-- TODO right sidebar gets too long, we should use tabs here -->
       <h3 id="{{ $file }}">
-        {{ $tool }} with <code>{{ $sampling }}</code> sampling type
+        {{ $tool }} with <code>{{ $sampling }}</code> sampling
```
albertteoh, line 99: Before: "... with adaptive sampling type"
After: "... with adaptive sampling"

scripts/gen-cli-data.py
```diff
@@ -36,24 +53,26 @@ def generate(tool, storage='', sampling=''):
...
-        )
+            f'{output_path}/data/cli/{jaeger_ver}/{tool}.yaml',
+            f'{output_path}/data/cli/{jaeger_ver}/{tool}-{feature_name}.yaml')
 
 
 for tool in cfg['tools']:
     tool_cfg = cfg[tool]
-    storage_types=tool_cfg['storage']
-    sampling_types=tool_cfg['sampling']
-    if storage_types:
```
albertteoh, line 50: These `if`s were redundant as a "zero" value of a list just means the body of the `for` loop won't be executed.

scripts/gen-cli-data.py
```diff
@@ -3,31 +3,48 @@
...
-    print('Generating YAML for {}, storage {} and sampling {}'.format(tool, storage, sampling))
 
-    volume='%s/data/cli/%s:/data' % (output_path, jaeger_ver)
-    docker_img='all-in-one' if tool == 'jaeger-all-in-one' else tool
-    docker_image="jaegertracing/%s:%s" % (docker_img, jaeger_ver)
-    storageOrSampling=storage or sampling
+# generate generates the CLI documentation for a given "tool" (e.g.
+# jaeger-collector, jaeger-query, etc.) and feature (e.g. storage=elasticsearch,
+# sampling=adaptive, metrics-storage=prometheus).
+def generate(tool, **kwargs):
```
albertteoh, line 20: I think using `kwargs` would be simpler and cleaner than adding another "optional" parameter to support metrics storage, especially avoiding the need to extend the `storageOrSampling` variable.

yurishkuro:
I think we need to rethink the organization of this doc. Each entry repeats a lot of options not related to the given storage type. It was sort of ok when there was only one dimension we could vary, now we have three.

What if we had a "core flags" section first and then each storage type would only list additional flags it enables? Could we isolate those extra flags by a common prefix?

jpkrohling:
> I think we need to rethink the organization of this doc. 

For this PR, or for the future? Should I wait on this one to release the new docs for 1.29.0?

yurishkuro:
> For this PR, or for the future? Should I wait on this one to release the new docs for 1.29.0?

Well, I guess since we're right in the middle of the release we should merge this as is and refactor later.

jpkrohling (approved):

albertteoh:
> I think we need to rethink the organization of this doc. Each entry repeats a lot of options not related to the given storage type. It was sort of ok when there was only one dimension we could vary, now we have three.
> 
> What if we had a "core flags" section first and then each storage type would only list additional flags it enables? Could we isolate those extra flags by a common prefix?

I agree that the page has a lot of duplicated flags; I'd originally thought that was by design so all possible flags are localised to that particular configuration to avoid jumping around different sections of the document.

I've created a follow-up ticket to discuss and/or address this concern: #541

PR #546, opened 2022-01-08
Rename master to main

albertteoh:
- Renames references from master to main in an attempt to fix the [failing release build job](https://github.com/jaegertracing/documentation/runs/4745349895?check_suite_focus=true) on the newly introduced `main` branch.

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -2,7 +2,7 @@ name: Release
 
 on:
   create:
-    branches: [ master ]
+    branches: [ main ]
     tags: ["release-*.*.*"]
 
 jobs:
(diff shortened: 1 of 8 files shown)


albertteoh:
> in an attempt to fix the failing release build job on the newly introduced main branch.

Hmmm... maybe this is expected behaviour based on my understanding of the [release script](https://github.com/jaegertracing/documentation/blob/main/scripts/release.sh#L23); in any case, we should still rename master to main.

yurishkuro (approved):
We may also need to update netlify settings

albertteoh:
> We may also need to update netlify settings

How can I do this? I checked https://github.com/jaegertracing/documentation/README.md (and CONTRIBUTING.md), but couldn't see instructions on how this is done. I also checked our maintainers-only secrets but couldn't find netlify credentials there.

I created my own netlify account and managed to configure it to deploy off `main`, but I'm guessing we need to login with a shared jaeger account to modify those settings.

yurishkuro:
We don't use shared credentials on Netlify, it has a team, I sent you invite to that team.

Meanwhile, I changed the deploy setting to pull from `main` branch and triggered a deploy manually, looks good.

PR #566, opened 2022-03-24
Fixing ILM documentation 

rbizos:
**Problem**

By following the documentation, the elasticsearch index template would be overridden by the collector. The template create by the collector would not have the ILM policy setup thus stopping the rollover to work.

**Proposal** 
Adding create-index-templates=false will prevent this issue. Having this mentioned in the doc will help people setup ILM without having to debug why the rollover would stop working.

### DIFF
--- a/content/docs/1.22/deployment.md
+++ b/content/docs/1.22/deployment.md
@@ -406,7 +406,7 @@ For example:
   "ILM policy jaeger-ilm-policy doesn't exist in Elasticsearch. Please create it and rerun init"
   {{< /info >}}
 
-  After the initialization, deploy Jaeger with `--es.use-ilm=true` and `--es.use-aliases=true`.
+  After the initialization, deploy Jaeger with `--es.use-ilm=true`, `--es.use-aliases=true` and `--es.create-index-templates=false`.
 
 
 #### Upgrade Elasticsearch version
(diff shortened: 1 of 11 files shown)


rbizos:
Another possibility I see is this: https://github.com/rbizos/jaeger/commit/124492197a7401b68e6e2bc45c44149ccf9025df

yurishkuro:
> Another possibility I see is this: https://github.com/rbizos/jaeger/commit/124492197a7401b68e6e2bc45c44149ccf9025df

I like this better as an automatic fix rather than documentation / convention

rbizos:
I opened a PR on jaeger https://github.com/jaegertracing/jaeger/pull/3599

joe-elliott:
Since 1.33 is technically released I'm going to go ahead and cut the docs. If there are adjustments to be made we can make them to `next-release` and `1.33` easily enough.

yurishkuro:
Closing, feel free to reopen.

albertteoh:
Yes, this can be closed. The changes in the PR addresses this issue without needing explicit documentation changes.

rbizos:
Hello, sorry for the delay. I would still have merged this change for version prior to 1.33 as the template creation needs to be explicitly disabled and could lead to surprises. After 1.33 it's not needed as mentioned :)

PR #568, opened 2022-03-29
Removal of a comment of the ES client for Jaeger ES

phoebe-canva:
Signed-off-by: Phoebe Zhou <phoebe@canva.com>

## Context
- Bulk.flush-interval is by default set to a value of 200ms
- This setting is only disabled if it it set to 0 but the last line of the documentation: "By default, this is disabled", is very misleading because by default, it is enabled with a value of 200ms

- Reduces confusion since there is a default value (but the comment says that it this setting is disabled)

- Removed the line "By default, this is disabled." to reduce confusion for users

### DIFF
--- a/data/cli/1.32/jaeger-ingester-elasticsearch.yaml
+++ b/data/cli/1.32/jaeger-ingester-elasticsearch.yaml
@@ -24,7 +24,7 @@ options:
 - name: es-archive.bulk.flush-interval
   default_value: 200ms
   usage: |
-    A time.Duration after which bulk requests are committed, regardless of other thresholds. Set to zero to disable. By default, this is disabled.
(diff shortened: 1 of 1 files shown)


yurishkuro:
Thanks, but these yaml files are generated automatically by running `$binary docs --format yaml`, we do not edit them manually. 

It's not clear to me why the comment string says "disabled by default" when the default value is set to 200ms, not zero:
```
$ SPAN_STORAGE_TYPE=elasticsearch go run ./cmd/collector help | grep es.bulk.flush

      --es.bulk.flush-interval duration                           A time.Duration after which bulk requests are committed, regardless of other thresholds. Set to zero to disable. By default, this is disabled. (default 200ms)
```

PR #572, opened 2022-05-10
Add more troubleshooting tips including metrics

albertteoh:
Signed-off-by: albertteoh <see.kwang.teoh@gmail.com>

- Resolves https://github.com/jaegertracing/jaeger/issues/3587

- With metrics available on the metrics reader, this adds troubleshooting tips that incorporate inspecting jaeger metrics.
- Additional troubleshooting advice, by following the flow of data up to its source, is also provided, along with relevant links.

### DIFF
--- a/content/docs/next-release/spm.md
+++ b/content/docs/next-release/spm.md
@@ -79,7 +79,7 @@ graph
     UI[Jaeger UI] --> QUERY
     QUERY[Jaeger Query Service] --> METRICS_STORE[Metrics Storage]
     PROMETHEUS_EXPORTER --> |metrics| METRICS_STORE
-    subgraph Opentelemetry Collector
+    subgraph OpenTelemetry Collector
         subgraph Pipeline
             TRACE_RECEIVER
             SPANMETRICS_PROC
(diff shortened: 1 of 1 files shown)


content/docs/next-release/spm.md
```diff
@@ -172,6 +172,93 @@ the HTTP API.
...
+jaeger_query_requests_total{operation="get_error_rates",result="err"} 4
+jaeger_query_requests_total{operation="get_latencies",result="err"} 8
+```
+
+At this point, checking the logs will provide more insight towards root causing
+the problem.
+
+### Query Prometheus
+
+Graphs may still appear empty even when the above jaeger metrics indicate successful reads
```
yurishkuro, line 241:
```suggestion
Graphs may still appear empty even when the above Jaeger metrics indicate successful reads
```

I this because the reads for non-existing metrics still succeed in Prometheus? Is it possible to detect that elsewhere, e.g. by emitting another metric that counts the number of data points returned from Prom?

albertteoh: > I (sic) this because the reads for non-existing metrics still succeed in Prometheus? 

Yup: `curl http://localhost:9090/api/v1/query?query=doesnotexist` will return HTTP status code `200` and an empty result.

> Is it possible to detect that elsewhere, e.g. by emitting another metric that counts the number of data points returned from Prom?

Ah, so perhaps this is what the [`response` counter](https://github.com/jaegertracing/jaeger/pull/3667#discussion_r868088359) was for, to build up a histogram on the number of results returned.

I'll create a follow-up PR for that.

yurishkuro (approved):

Issue #570, opened 2022-04-22
Home page load is too slow


yurishkuro:
**Describe the bug**
On my laptop the home page loads to the state where it is already rendered on the screen, but it's not interactive until about 8sec later.

**Additional context**
Running a profiler showed that between 2.5s and 8s there are no URL loading, instead all the time is spend in "scripting" in the functions of lunr library, which we use for searching documents. As I understand, it parses the 2.4Mb index.json from the root into some internal index.

The following diff changes that index.json to only include the latest version:
```diff
diff --git a/themes/jaeger-docs/layouts/index.json b/themes/jaeger-docs/layouts/index.json
index f29843c..fc7890b 100644
--- a/themes/jaeger-docs/layouts/index.json
+++ b/themes/jaeger-docs/layouts/index.json
@@ -1,7 +1,10 @@
+{{ $latestVersion := site.Params.latest }}
 {{ $index := newScratch }}
 {{ $index.Add "index" slice }}
 {{ range where site.Pages "Section" "docs" }}
-{{ $version := index (split .File.Path "/") 1 }}
-{{ $index.Add "index" (dict "title" .Title "url" .Permalink "body" (.Content | plainify | jsonify) "summary" .Summary "version" $version) }}
+  {{ $version := index (split .File.Path "/") 1 }}
+  {{ if eq $version $latestVersion }}
+    {{ $index.Add "index" (dict "title" .Title "url" .Permalink "body" (.Content | plainify | jsonify) "summary" .Summary "version" $version) }}
+  {{ end }}
 {{ end }}
 {{ $index.Get "index" | jsonify }}
```

This could be a stop-gap solution, but it's not ideal because the Search bar actually only shows when in the Docs, so it does make sense for that search to work within the documents of a specific version. I think a better solution would be
  * remove lunr from the home page and other static pages, only show it when in the Docs pages
  * generate multiple index.json files for each `doc/{$version}` independently and only load that index for pages in that version

PR #575, opened 2022-05-13
Disable search on all docs except latest version

yurishkuro:
Temporary workaround for #570:
  * do not include non-latest docs into the search index `index.json`
  * do not render the Search bar on non-latest docs
  * do not load lunr module on non-docs pages

cc @albertteoh

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -1,7 +1,7 @@
 name: Release
 
 on:
-  create:
+  push:
     branches: [ main ]
     tags: ["release-*.*.*"]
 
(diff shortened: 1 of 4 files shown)


albertteoh (approved):

.github/workflows/ci-release.yml
```diff
@@ -1,7 +1,7 @@
 name: Release
 
 on:
-  create:
+  push:
```
albertteoh, line 4: For my learning, what does changing `create`->`push` fix?

yurishkuro: I don't think create was a valid entry at all, hence the workflow kept running and failing on all PRs (because we check for tag in the script).

PR #582, opened 2022-07-11
added description about custom CNI issue

klubi:
Signed-off-by: Piotr Klubinski <piotr.klubinski@jamf.com>

Describes issue faced in [jaeger-operator](https://github.com/jaegertracing/jaeger-operator/issues/1963)

Described workaround for issue with deploying jaeger on AWS EKS clusters running custom CNI (Calico)


klubi:
Exact preview of what was added is here -> https://deploy-preview-582--romantic-neumann-1959d7.netlify.app/docs/next-release/operator/#calico-cni

yurishkuro:
This feels very specific to AWS. I suggest instead creating a Discussion question in the main repo and posting the answer, as a form of knowledge base.

klubi:
Correct, it's AWS specific (both EKS and Fargate deployments will be impacted).
I've seen however OpenShift specific info regarding operator deployment, so I thought it would be appropriate to put it there.
Cert-manager has this info i troubleshooting section -> https://cert-manager.io/docs/concepts/webhook/#webhook-connection-problems-on-aws-eks, maybe Jaegers Troubleshooting section will be the right place?

tigerquoll:
The workaround being documented is relevant for any AWS EKS (at least) deployment where the master nodes are on different CNI's then the working nodes. EKS users cannot modify the CNI that the master node is deployed on, so any change to what CNI the worker nodes use means you have to surface ports on the host network to allow for any non standard k8s comms between them.
BTW:  Webhook-bind-port appears to be an integer so ":10290" is not valid - "10290" would be

PR #603, opened 2023-02-10
Add test workflow

yurishkuro:
* Add test workflow that runs the build and checks local links

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,13 +19,14 @@ develop:	generate
 	HUGO_PREVIEW=true hugo server \
         --buildDrafts \
         --buildFuture \
-	--ignoreCache
+		--ignoreCache
 
 clean:
 	rm -rf public
 
 netlify-production-build:	generate
 	hugo --minify
+	rm -rf public/_client_libs
 
 netlify-deploy-preview:	generate
 	HUGO_PREVIEW=true hugo \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -19,30 +19,34 @@ develop:	generate
 	HUGO_PREVIEW=true hugo server \
         --buildDrafts \
         --buildFuture \
-	--ignoreCache
```
albertteoh, line 22: nit: extra indent intentional?

Makefile
```diff
@@ -19,30 +19,34 @@ develop:	generate
...
+		--ignoreCache
 
 clean:
 	rm -rf public
 
 netlify-production-build:	generate
 	hugo --minify
+	rm -rf public/_client_libs
```
albertteoh, line 29: Can we DRY these up by placing it under the `generate` make target?

Makefile
yurishkuro: e1fd402

Makefile
albertteoh: Ah, please ignore this comment 😄

yurishkuro: No, this needs to run after the hugo build, and hugo commands are all different. Only way to DRY is to change to something like `$(MAKE) unpublish_client_libs`

I actually tried to use Hugo's `ignorefiles` directive in the config, e.g. 
```
ignorefiles: [ "content/_client_libs/*md" ]
```
but it ends up completely removing (not generating) the respective files under `content/docs/$ver/`, even though their paths do not match these ignored files. I suspect Hugo performs symlink to absolute path  resolution (for whatever reason), it's the only way it could be matching on the file names in a different location.

PR #623, opened 2023-04-05
add a note about scylladb as a storage backend to deployment.md

mkorolyov:

This PR solves the lack of documentation regarding the use of ScyllaDB as a drop-in replacement for Cassandra in Jaeger tracing server deployment. By providing clear instructions on how to configure the CASSANDRA_SERVERS environment variable, users can easily switch to ScyllaDB for enhanced performance and compatibility without any additional changes to their existing setup.

This PR adds a note to the Jaeger tracing server deployment documentation, explaining how to configure the system to use ScyllaDB as a drop-in replacement for Cassandra by simply updating the CASSANDRA_SERVERS environment variable to point to ScyllaDB hosts.

### DIFF
--- a/content/docs/next-release/deployment.md
+++ b/content/docs/next-release/deployment.md
@@ -472,6 +472,22 @@ usercert = ~/.cassandra/client-cert
 # validate = false
 ```
 
+### ScyllaDB
+Last checked version: 5.1.7
+
+ScyllaDB can be utilized as a drop-in replacement for Cassandra when deploying the Jaeger tracing server. The integration process is straightforward and requires minimal changes to your existing configuration.
+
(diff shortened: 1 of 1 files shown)


content/docs/next-release/deployment.md
```diff
@@ -472,6 +472,22 @@ usercert = ~/.cassandra/client-cert
 # validate = false
 ```
 
+### ScyllaDB
```
yurishkuro, line 475: Since ScyllaDB is not an officially supported backend, it's misleading to include it in this list.

We can add an L3 Section after Plugins called "Compatible Backends" and have a simple bullet list that includes ScyllaDB, with a link to the README in the main repo with Docker example.

mkorolyov: thanks, updated

yurishkuro (approved):

PR #629, opened 2023-04-14
Consistent Naming For Jaeger Components

### LINKED ISSUE #612: [Feature]: Use consistent naming for Jaeger components
### Requirement

Documentation should sue consistent naming to refer to Jaeger components.

### Problem

Some pages are using the binary names like `**jaeger-ingester**`, which is our preferred way, but others still use capitalized component names, like "Jaeger Agent", "the Jaeger Agent".

### Proposal

Update all pages to the same nomenclature:
  * Always use the binary name, such as `**jaeger-agent**`, `**jaeger-collector**`, `**jaeger-query**`
  * Do not prefix them with "the"
  * Make plural forms by adding `s` after the bold portion, i.e. `**jaeger-collector**s` == **jaeger-collector**s
  * Similarly, make possessive forms by adding `'s` after the bold portion, i.e. `**jaeger-collector**'s` == **jaeger-collector**'s
  * The only exception is when the section header refers to just the component, as in Deployment page that has headers like `Collector`, `Agent`, etc. While we could change those too, this will break the existing hyperlinks, so it's not worth it.

Incorrect usages can be located with a grep:
```shell
grep -rine '[^-]agent' content/docs/next-release
grep -rine '[^-]collector' content/docs/next-release
grep -rine '[^-]query' content/docs/next-release
grep -rine '[^-]ingester' content/docs/next-release
```

### Open questions

_No response_

SaarthakMaini:
Fixes #612

Naming Convention Of:

- apis.md
- deployment.md
- faq.md
- getting-started.md

made consistent with the guidelines given in the issue

### DIFF
--- a/content/docs/next-release/apis.md
+++ b/content/docs/next-release/apis.md
@@ -11,33 +11,33 @@ The following labels are used to describe API compatibility guarantees.
 * **internal** - the APIs intended for internal communications between Jaeger components and not recommended for use by external components.
 * **deprecated** - the APIs that are only maintained for legacy reasons and will be phased out in the future.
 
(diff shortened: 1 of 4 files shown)


yurishkuro (approved):
Fantastic!

PR #645, opened 2023-07-13
document the query api so it could be used externally

yuvalif:
- query api is not well documneted

- added missign documentation


yurishkuro:
It doesn't matter if it's stable, we don't want to put a restriction on it to be stable & backwards compatible, we want to be able to change it as needed without affecting users. The officially supported API is gRPC.

PR #646, opened 2023-08-12
Documentation Changes for Critical Path Analysis

GLVSKiriti:
This PR is to add documentation for critical path analysis to Jaeger UI.
Resolves  https://github.com/jaegertracing/jaeger-ui/issues/1288
@yurishkuro  @yashrsharma44


GLVSKiriti:
Should I make a separate section to explain about Critical Path features like when in collapsed state it shows combined CP etc...
or should I add them in features or introduction section.
Present I replaced the images in which critical path of each span is shown

yurishkuro:
I am tempted to remove these embedded mode screenshots altogether, doesn't feel like the provide a lot of value - people always can try it on a live UI with the provided URLs.

Issue #665, opened 2024-01-19
Add spell checker to CI


yurishkuro:
Something similar to github.com/open-telemetry/opentelemetry-collector/pull/9318

jaydee029:
Hi, I would like to work on this.

h4shk4t:
Hi @yurishkuro I would like to work on this issue.

PR #668, opened 2024-02-01
Add Spell Checker to CI

h4shk4t:
- Closes [#665](https://github.com/jaegertracing/documentation/issues/665)

## Description of the changes
- Simply added a new cspell job to the existing ci-test workflow.
- Will perform the spell check in the content directory and its subdirectories.
- Added the same set of dictionary as suggested in the reference PR in the issue thread. (Let me know if any more words need to be added)


yurishkuro (commented):
doesn't look like it did anything in the CI run

.github/workflows/ci-test.yml
```diff
@@ -29,3 +29,10 @@ jobs:
     - name: Build
       run: |
         make build
+    
+    - name: Spelling Check
+      uses: streetsidesoftware/cspell-action@v5.3.0
```
yurishkuro, line 34: I don't like using magic GH actions for linters. It should be possible to run the linter locally, e.g. `make spellcheck`

h4shk4t: Alright understood, I will make the changes and update again

.project-words.txt
```diff
@@ -0,0 +1,143 @@
+accountingservice
```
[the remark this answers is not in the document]
h4shk4t: From the PR you shared on the GitHub issue thread, it seemed to have a good set of words which I thought would apply here as well as mentioned in the PR description. Should I manually add the words instead?

.github/workflows/ci-test.yml
h4shk4t: Wrote a python for spell check. Could you review?

.project-words.txt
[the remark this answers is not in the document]
h4shk4t: I ran the script once and stored all the misspelled words in the first iteration as the words to ignore. Kindly check if they are sufficient. It is placed in scipts/utils folder.

scripts/utils/dictionary.txt
```diff
@@ -0,0 +1,723 @@
+aspecto
```
yurishkuro, line 1: this file doesn't make sense to me. E.g. why are commit hashes in there, or things like p50 or src? If the checker is so dumb as to flag every single string, then there's no value in it for us, it's always going to be too noisy.

h4shk4t: Hmm, i get your point. Could you recommend a local spellchecker that runs well? I'll use that instead

yurishkuro: I cannot, research is part of the task. I think OTEL website used one, other CNCF websites like Kubernetes might use one too.

h4shk4t: So I researched a few spelling checker libraries and found 2 potential candidates. One is GNU Aspell and the other is [misspell](https://pkg.go.dev/github.com/client9/misspell/cmd/misspell) by client9. I tested both of them and I found Aspell to perform better at detecting wrong spellings and ignoring false positives like hexadecimal strings. This leaves us with 4 options (if we include the old options as well): Aspell, misspell, cspell (GitHub action), and pyspellchecker (in decreasing order of preference). I can continue looking for more spellcheckers if these are not satisfactory.

h4shk4t: I wrote this implementation in Aspell. I created a dictionary on the basis of the initial run of Aspell over the contents. I tried my best to not accidentally add actual typos or misspellings generated from the command while writing the dictionary. One problem I encountered was of encoding of words like `Kröhling` and `Paixão`. Aspell was struggling to deal with the unicode characters of these words despite setting the encoding to UTF-8. The alternative was to use Hunspell, but it performs worse overall when it comes to checking spellings. Apart from that, the spellcheck seems to be working fine.

yurishkuro: > words like Kröhling and Paixão

does it not work if we add these words to the dictionary?

my_dictionary.txt
```diff
@@ -0,0 +1,289 @@
...
+autoscale
+autoscaler
+backend
+backends
+bak
+balancer
+bansal
+bohara
+bootcamp
+cassandra
```
yurishkuro, line 27: what kind of standard dictionary are you using? I would've expected these words to be pre-defined somewhere.

my_dictionary.txt
```diff
@@ -0,0 +1,289 @@
...
+logz
+lookback
+loopback
+lyft
+macos
+mailgroup
+marck
+mentee
+mentees
+mentorship
```
yurishkuro, line 150: mentorship is not a recognized word?

my_dictionary.txt
h4shk4t: I was not using a standard dictionary, what I did was run the spellchecker once and I manually added all the words from the output which were misclassified as incorrect spellings but were technical terms or proper nouns. Should I also add more words from another source?

my_dictionary.txt
h4shk4t: Even I found that to be odd that Aspell's default dictionary doesnt recognise mentorship and some other words. However, when I tried alternatives like misspelling or hunspell I faced a lot more other issues like reporting hashes and hexadecimal strings as wrong spellings, etc. As Aspell has been the best in this regard I thought I'd just add it to the dictionary to make it work (there are not that many words with such false positives as far as I know)

yurishkuro: There are standard dictionaries in aspell ecosystem, aren't there? We need to use those and only add our dictionary with our specific terms.

h4shk4t: Yes this dictionary is on top of the standard dictionary offered by the aspell ecosystem.

scripts/utils/dictionary.txt
h4shk4t: On running the script I get the following error (after adding these words to the dictionary) 
image

yurishkuro:
The spellchecker flags normal words, so I don't see it as a viable solution at this point. Maybe it needs better dictionaries. Also, I would argue it should not be checking test formatted as code, because that will always bring a lot of noise.

h4shk4t:
@yurishkuro I did some digging and found the latest dictionary by `aspell` from 2020 in one of their mirrors. It shows some improvement where it is not flagging words like `mentorship`, `analytics` and more (verified this manually). However there are still a lot more common words which are still being flagged like `roadmap` and `onboarding`. I was trying Hunspell again, but I am facing issues with making it ignore hexadecimal values (will try to look for some bypasses to fix this issue in the meantime). Kindly advise if I should continue with `aspell` with a custom dictionary which also includes words like roadmap and onboarding, or to try and make `Hunspell` reduce false positives from code blocks and hexadecimal values.

yurishkuro:
@h4shk4t I don't have preference in the tool, I only have preference for the end result - low friction :-)

h4shk4t:
@yurishkuro Made the changes. Using `cspell` binary for spellchecking. Gives a promising accuracy and far lesser tolerance for false positives. Kindly let me know if this suffices. 

Best Regards

project-words.txt
```diff
@@ -0,0 +1,166 @@
+hasparent
```
yurishkuro, line 1: Please sort and add a linter check that it is sorted.

Is this list case sensitive?

scripts/requirements.txt
```diff
@@ -0,0 +1,2 @@
+pyspellchecker
+markdown-it-py
```
yurishkuro, line 2: is this needed?

cspell.json
```diff
@@ -0,0 +1,12 @@
+{
```
yurishkuro, line 1: adding a lot of files to the root dir, can we stash them all somewhere, like `scripts/cspell/*`?

.github/workflows/ci-test.yml
```diff
@@ -28,4 +28,4 @@ jobs:
 
     - name: Build
       run: |
-        make build
+        make build
```
yurishkuro, line 31: revert the change

project-words.txt
h4shk4t: Made the fixes

scripts/requirements.txt
h4shk4t: Made the fixes

.github/workflows/ci-test.yml
h4shk4t: Reverted the change

cspell.json
h4shk4t: Made the fixes

.github/workflows/ci-test.yml
```diff
@@ -27,5 +27,4 @@ jobs:
         hugo version
 
     - name: Build
-      run: |
-        make build
+      run: |
```
yurishkuro, line 30: why do you need to change this file?

scripts/cspell/spellcheck.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+if ! sort -c project-words.txt; then \
+    echo "project-words.txt is not sorted." && exit 1; \
+fi
+npm install cspell
+npm run spellcheck
```
yurishkuro, line 7: in which directory will it run?

scripts/cspell/spellcheck.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+if ! sort -c project-words.txt; then \
+    echo "project-words.txt is not sorted." && exit 1; \
+fi
```
yurishkuro, line 5:
```suggestion
if ! sort -c project-words.txt; then
    echo "project-words.txt is not sorted."
    exit 1
fi
```

Makefile
```diff
@@ -58,3 +58,6 @@ check-internal-links: clean build link-checker-setup run-link-checker
 
 check-all-links: clean build link-checker-setup
 	bin/htmltest --conf .htmltest.external.yml
+
+spellcheck:
```
yurishkuro, line 62: this is not called from CI anywhere

.github/workflows/ci-test.yml
h4shk4t: I misunderstood what was asked. I reverted the change to the file by reverting back to the commit `Disable links checking...` on the ci-test.yml file.

scripts/cspell/spellcheck.sh
h4shk4t: It is being called from the `scripts/cspell/` directory. But the spellcheck script in `package.json` traverses to the content folder and performs a spellcheck on all the markdown files inside it.

scripts/cspell/spellcheck.sh
h4shk4t: Made the fixes

Makefile
h4shk4t: Now `ci-test.yml` is calling spellcheck and exits if a misspelling is found.

yurishkuro:
The lint step is catching some real misspellings now, plus a few words that need to go to the dictionary

h4shk4t:
> The lint step is catching some real misspellings now, plus a few words that need to go to the dictionary

Yes, I will add the new words to the dictionary (must have missed them earlier). What should I do about the misspellings?

yurishkuro (approved):

PR #676, opened 2024-03-03
List all storage backends are currently supported for adaptive sampling

Pushkarm029:
- part of https://github.com/jaegertracing/jaeger/issues/3305

## Description of the changes
- I have updated the document to include information about storage backends that support sampling.

## Checklist

### DIFF
--- a/content/docs/next-release/sampling.md
+++ b/content/docs/next-release/sampling.md
@@ -92,7 +92,7 @@ Since Jaeger v1.27.
 
 Adaptive sampling works in **jaeger-collector** by observing the spans received from services and recalculating sampling probabilities for each service/endpoint combination to ensure that the volume of collected traces matches `--sampling.target-samples-per-second`. When a new service or endpoint is detected, it is initially sampled with `--sampling.initial-sampling-probability` until enough data is collected to calculate the rate appropriate for the traffic going through the endpoint.
 
-Adaptive sampling requires a storage backend to store the observed traffic data and computed probabilities. At the moment `memory` (for all-in-one deployment) and `cassandra` are supported as sampling storage backends. We are seeking help in implementing support for other backends ([tracking issue](https://github.com/jaegertracing/jaeger/issues/3305)).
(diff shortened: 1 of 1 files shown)


content/docs/next-release/sampling.md
```diff
@@ -92,7 +92,7 @@ Since Jaeger v1.27.
 
 Adaptive sampling works in **jaeger-collector** by observing the spans received from services and recalculating sampling probabilities for each service/endpoint combination to ensure that the volume of collected traces matches `--sampling.target-samples-per-second`. When a new service or endpoint is detected, it is initially sampled with `--sampling.initial-sampling-probability` until enough data is collected to calculate the rate appropriate for the traffic going through the endpoint.
 
-Adaptive sampling requires a storage backend to store the observed traffic data and computed probabilities. At the moment `memory` (for all-in-one deployment) and `cassandra` are supported as sampling storage backends. We are seeking help in implementing support for other backends ([tracking issue](https://github.com/jaegertracing/jaeger/issues/3305)).
+Adaptive sampling requires a storage backend to store the observed traffic data and computed probabilities. At the moment `memory` (for all-in-one deployment), `cassandra`, `badger` and `elasticsearch` are supported as sampling storage backends.
```
yurishkuro, line 95: Since elasticsearch and opensearch are currently implemented with the same plugin in Jaeger, does this mean opensearch can also be used for adaptive sampling?

Pushkarm029: I think so because when adaptive sampling was messed up, integration tests failed for both open search and elastic search. So, we can conclude that adaptive sampling will also work with OpenSearch.

If you agree too, we will add it here(in docs), also

yurishkuro: +1 - we don't have specific opensearch code in `plugin/storage/integration` which means it's working for it too

yurishkuro:
Cannot amend PR from fork/main, resubmitted as #688

PR #684, opened 2024-03-14
Add jaeger-postgresql to storage plugins list

robbert229:
This PR lists the postgres-jaeger storage plugin that I maintain to the storage plugins list.

## Description of the changes
Adds a link to the storage plugin list.

It was visually checked.
image

## Checklist
- [X] I have added unit tests for the new functionality
- [X] I have run lint and test steps successfully

### DIFF
--- a/content/docs/next-release/deployment.md
+++ b/content/docs/next-release/deployment.md
@@ -760,7 +760,7 @@ Available sidecar plugins:
 * [InfluxDB](https://github.com/influxdata/influxdb-observability/blob/main/jaeger-influxdb/README.md) - time series database.
 * [Logz.io](https://github.com/logzio/jaeger-logzio) - secure, scalable, managed, cloud-based ELK storage.
 * [ClickHouse](https://github.com/jaegertracing/jaeger-clickhouse) - fast open-source OLAP DBMS.
-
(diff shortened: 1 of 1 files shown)


jkowall (approved):
Thanks for your contribution and the plugin, have you considered how this will work with Jaeger v2 with the new collector?

robbert229:
Good question, as soon as v2 gets to a mature state I plan on making the migration immediately.

yurishkuro:
@robbert229 is it implemented as a go-plugin or gRPC server? We plan to sunset the go-plugin support and only support remote servers implementing Remote Storage API (which is mostly how go-plugin also works plus some sidecar mechanics). In Jaeger-v2 the gRPC path is already supported (but we also plan to introduce a V2 Storage API that will be batch-based, not single span like V1 API).

robbert229:
It's implemented as a gRPC Server. 

I will say that the gRPC server experience is significantly better in my opinion than the original plug-in developer experience.

yurishkuro:
> I will say that the gRPC server experience is significantly better in my opinion than the original plug-in developer experience.

this is why we want to sunset the latter :-). https://github.com/jaegertracing/jaeger/issues/4647

PR #690, opened 2024-04-03
Fix: Navbar hover background color much darker

### LINKED ISSUE #658: [Feature]: Use darker background for topnav menu mouseover
### Requirement

It is a part of UI enhancement.

### Problem

Currently, when we hover over an element, its background changes to white, specifically the header's background, which doesn't create a very visible effect. To enhance the webpage's UI, it should be more prominently visible when we hover over a navigation element.

Screenshort:
image

### Proposal

When a user hover an Nav element it should change it color from black to color: #66cfe3;(primary color).
or 
Suggest me any other enhancement that can i do.

### Open questions

_No response_

tico88612:
- Resolves #658

## Description of the changes
- Navbar hover background color much darker

- manual test

## Checklist


themes/jaeger-docs/assets/sass/navbar.sass
```diff
@@ -14,4 +14,7 @@
 
       &.is-small
         padding: 0 0 0 2rem
-        font-size: 1.1rem
\ No newline at end of file
+        font-size: 1.1rem
+
+      &:hover
+        background-color: $navbar-item-hover-background-color
```
yurishkuro, line 20: `$navbar-item-hover-background-color` was already previously defined and the color **was** already changing to light grey. So there had to be a style defined that used that - why do we need a new one?

tico88612: The original way of writing would cause the background color of the dropdown item to be inconsistent when hovering because this line was absent.

But after looking into it again, I think I should change the `$navbar-dropdown-item-hover-background-color` variable instead of working on the CSS myself.

yurishkuro:
Do you think we can also add this highlighting to the left and right sidebar menus?

image

tico88612:
Does it look like this? (only left and right sidebar)

https://github.com/jaegertracing/documentation/assets/17496418/a127787e-b536-4a1c-ab1f-98a3f371b575

yurishkuro:
Not quite - this changes all links, I was only thinking of the menus and the highlights would be row level, not text-only, similar to how it's in the top nav bar

tico88612:
Disclaimer: I'm not a UI/UX designer; I'm just examining it from my own aesthetic point of view.
---
The right-side menu is OK.
Does the left menu look like this movie if it is changed to look like the navigation bar above? (from bulma.io)

https://github.com/jaegertracing/documentation/assets/17496418/27a2375b-388d-4093-a166-49b0b490bb7c

But we can open a new issue to discuss this.

yurishkuro:
you're right, the right side menu is already using row highlighting, only the left one doesn't. The 2nd video looks great as an example (it even does indentation of the highlight on children menu items)

yurishkuro (approved):

PR #694, opened 2024-04-19
Fixing a small typo

virtualandy:

Eek, I didn't create an issue since this was such a small change. I can definitely do that if needed!

## Description of the changes

Just fixing a typo.

I didn't run locally 😭 since just a typo fix.

## Checklist

### DIFF
--- a/scripts/cspell/project-words.txt
+++ b/scripts/cspell/project-words.txt
@@ -23,7 +23,7 @@ datacenter
 datasource
 developercertificate
 dilligence
-distirbuted
+distributed
 documentiing
 downsampling
 dropwizard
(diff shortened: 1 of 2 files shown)


virtualandy:
The typo is in versions 1.56 and earlier - if I got this wrong by editing `next-release` let me know and I'm happy to fix it up!

yurishkuro (commented):
thanks for catching. Turns out we have an invalid word in the allowed dictionary, and not the only one wrong. We should fix if there first and the linter will flag all occurances.
https://github.com/jaegertracing/documentation/blob/89637b6f45a80e2b4bef55d6d261557a40021345/scripts/cspell/project-words.txt#L26

virtualandy:
@yurishkuro oh, neat, didn't realize that was there (apologies, I should've read all the docs/setup flow first :) ).
I made an update to that word to see how it affects the build/tests. I'll scan it for other typos too.

scripts/cspell/project-words.txt
```diff
@@ -23,7 +23,7 @@ datacenter
 datasource
 developercertificate
 dilligence
-distirbuted
+distributed
```
yurishkuro, line 26: I don't believe we need this in the words list. btw dilligence is also misspelled (only needs one l)

virtualandy:
Looks like this was fixed with https://github.com/jaegertracing/documentation/pull/704

Issue #701, opened 2024-05-11
[Bug]: fix spelling issues


yurishkuro:
We spellcheck the content using `make spellcheck`, which uses a dictionary of words `scripts/cspell/project-words.txt`. However, this dictionary was seeded with some clearly misspelled words, at minimum those shown in the diff below. The goal is to fix the occurrences of those misspelled words in the main content (e.g. by finding them with `grep -rn {word} content/`) and remove them from the dictionary file.

```diff
--- a/scripts/cspell/project-words.txt
+++ b/scripts/cspell/project-words.txt
@@ -6,15 +6,12 @@ ashmita
 autoscale
 autoscaler
 bansal
-behaviour
 bohara
 bootcamp
-boths
 certfile
 changeme
 clickhouse
 cncf
-conrolled
 containerd
 cqlsh
 cqlshrc
@@ -22,9 +19,6 @@ daemonset
 datacenter
 datasource
 developercertificate
-dilligence
-distirbuted
-documentiing
 downsampling
 dropwizard
 dsoie
@@ -51,8 +45,6 @@ htpasswd
 ingester
 ingester's
 ingesters
-ingestor
-initialisation
 istio
 istio's
 jaegertracing
@@ -128,18 +120,15 @@ statefulset
 strimzi's
 struct
 tchannel
-telemtery
 thriftrw
 tolerations
 tracegen
-trft
 uberctx
 unmarshaling
 upsample
 usercert
 userkey
 valuen
-variery
 vlog
 volz
 weaveworks
```

priyanshsao:
does this issue is still present ? or has been solved.

priyanshsao:
the two words "telemtery","trft" are still on the file , does the words needs to be removed.

yurishkuro:
@priyanshsao yes they should be removed from the dictionaries

priyanshsao:
when i searched for the word "trft" i got these-

content/docs/1.16/operator.md:979:          name: jg-compact-trft
content/docs/1.17/operator.md:704:      name: zk-compact-trft
content/docs/1.17/operator.md:710:      name: jg-compact-trft
content/docs/1.17/operator.md:713:      name: jg-binary-trft
content/docs/1.17/operator.md:1145:          name: jg-compact-trft
content/docs/1.18/operator.md:720:      name: zk-compact-trft
content/docs/1.18/operator.md:726:      name: jg-compact-trft
content/docs/1.18/operator.md:729:      name: jg-binary-trft
[...]

PR #705, opened 2024-05-13
Fix: Typo for both, initialization, variety

JeevaRamanathan:
- Part of #701 

## Checklist


yurishkuro (approved):
Thanks!

PR #740, opened 2024-09-17
fixing padding and spacing to get TOC to render right

jkowall:
This change will fix the spacing on all of the docs to make the screen real state better with less white space. It makes
 a bigger difference in the v2 docs with the new TOC.

I did some testing on smaller screens as well.

Happy to get feedback.
- [X] I have run lint and test steps successfully

### DIFF
--- a/themes/jaeger-docs/assets/sass/pages.sass
+++ b/themes/jaeger-docs/assets/sass/pages.sass
@@ -57,8 +57,8 @@
       .nav
         position: absolute
         z-index: 10
-        width: 180px
-        max-width: 20rem
+        min-width: 15rem
+        max-width: 25rem
         line-height: 150%
         overflow-y: scroll
         overflow-x: hidden
--- a/themes/jaeger-docs/assets/sass/pages.sass
+++ b/themes/jaeger-docs/assets/sass/pages.sass
@@ -52,12 +52,13 @@
       position: fixed
       bottom: 0%
       top: 9rem
+      left: 3rem 
 
       .nav
         position: absolute
         z-index: 10
-        width: inherit
-        max-width: 11rem
+        width: 180px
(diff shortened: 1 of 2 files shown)


themes/jaeger-docs/assets/sass/pages.sass
```diff
@@ -52,12 +52,13 @@
       position: fixed
       bottom: 0%
       top: 9rem
+      left: 3rem
```
yurishkuro, line 55: I tried to play with this one and didn't understand what it affects, seems turning it off makes the sidebar actually larger

themes/jaeger-docs/assets/sass/pages.sass
```diff
@@ -52,12 +52,13 @@
...
       top: 9rem
+      left: 3rem 
 
       .nav
         position: absolute
         z-index: 10
-        width: inherit
-        max-width: 11rem
+        width: 180px
+        max-width: 20rem
```
yurishkuro, line 61: I tried this instead.
```suggestion
        min-width: 15rem
        max-width: 25rem
```
My version works better as Perf Tuning no longer gets wrapped. However neither setting works at 150% zoom or higher, the sidebar starts overlapping the main text (but that's probably the issue with how the main area is defined).

themes/jaeger-docs/assets/sass/pages.sass
jkowall: I added this to fix some of the strangeness in the scroll bar, but it depends on the browser you test with. If you don't think we need it I can remove it, just let me know.

yurishkuro: no, it's good, I see what it's doing.

yurishkuro (approved):

PR #747, opened 2024-09-23
Use full versions in Docker commands

yurishkuro:
- In v1.61.0 we stopped publishing shortcut Docker tags like `1` and `1.61`, and only publish fully qualified versions `1.61.0`
- But the docs were still referencing major/minor versions ([reported on Slack](https://cloud-native.slack.com/archives/CGG7NFUJ3/p1727056265543819))

## Description of the changes
- Use fully qualified versions everywhere
- The docs URLs still use major.minor, so rename the variables to `$majorMinorVersion` to avoid confusion with the `currentVersion` shortcut

### DIFF
--- a/themes/jaeger-docs/layouts/partials/docs/panel.html
+++ b/themes/jaeger-docs/layouts/partials/docs/panel.html
@@ -5,8 +5,8 @@
 {{ else }}
   {{ $path = .Path }}
 {{ end }}
-{{ $currentVersion := index (split $path "/") 1 }}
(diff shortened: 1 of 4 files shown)


Shawyeok:
@yurishkuro 
It seems new version variable included an extra newline character, some shell commands are syntax broken now, e.g.
image

yurishkuro:
on which page is it?

yurishkuro:
nvm, found it

yurishkuro:
Thanks, just pushed another fix a1b93ce

Issue #750, opened 2024-09-28
[jaeger-v2] Document changes to admin port / services


yurishkuro:
* metrics are on a different port `address: 0.0.0.0:8888` (`address: 0.0.0.0:8889` for ingester)
* expvar is also on a different port via `expvar` extension: `port: 27777`
* Healthchecks are on a different port via `healthcheckv2` extension: `endpoint: "0.0.0.0:13133"`

jkowall:
I did some updates to the docs per : https://github.com/jaegertracing/documentation/pull/751

The following seem to still be in the v1 docs and I believe should be somewhere in v2 docs, let me know what you think:

```
At default settings **jaeger-collector** exposes the following ports:

| Port  | Protocol | Endpoint | Function
| ----- | -------  | -------- | ----
| 4317  | gRPC     | n/a      | Accepts traces in [OpenTelemetry OTLP format][otlp] (Protobuf).
| 4318  | HTTP     | `/v1/traces` | Accepts traces in [OpenTelemetry OTLP format][otlp] (Protobuf and JSON).
| 14268 | HTTP     | `/api/sampling` | Serves sampling policies (see [Remote Sampling](../sampling/#remote-sampling)).
|       |          | `/api/traces` | Accepts spans in [jaeger.thrift][jaeger-thrift] format with `binary` thrift protocol (`POST`).
| 9411  | HTTP     | `/api/v1/spans` and `/api/v2/spans` | Accepts Zipkin spans in Thrift, JSON and Proto (disabled by default).
| 14250 | gRPC     | n/a      | Used by **jaeger-agent** to send spans in [model.proto][] Protobuf format.
```

yurishkuro:
yes, those ports would still be relevant (sometimes the ports might be different)

jkowall:
Is anything else pending on this issue or can we close @yurishkuro ?

yurishkuro:
I plan to give a read of all new docs this w/e and then will close.

PR #751, opened 2024-10-02
Replace command line args in deployment doc

jkowall:
Part of https://github.com/jaegertracing/documentation/issues/750

## Description of the changes
Adjusting command line and other components from v1 to v2.

## Checklist

### DIFF
--- a/content/docs/next-release-v2/deployment.md
+++ b/content/docs/next-release-v2/deployment.md
@@ -84,7 +84,7 @@ ElasticSearch:
           index_prefix: "jaeger-archive"
 ```
 
-The following ports can also be used by **jaeger** for varous types of integrations
+The following ports can also be used by **jaeger** for various types of integrations:
 
 | Port  | Protocol | Endpoint | Function
 | ----- | -------  | -------- | ----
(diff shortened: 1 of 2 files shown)


content/docs/next-release-v2/deployment.md
```diff
@@ -37,109 +36,55 @@ There are orchestration templates for running Jaeger with:
...
-```sh
-docker run \
-  --rm \
-  -p5775:5775/udp \
-  -p6831:6831/udp \
-  -p6832:6832/udp \
-  -p5778:5778/tcp \
-  jaegertracing/jaeger-agent:{{< currentVersion >}} \
-  --reporter.grpc.host-port=jaeger-collector.jaeger-infra.svc:14250
+  jaeger_storage:
```
yurishkuro, line 51: We have dedicated pages for each storage backend now, I would rather consolidate everything about that backend there, including config examples. This page could be just about different deployment roles, how they are different, the ports, the APIs exposed.

content/docs/next-release-v2/deployment.md
```diff
@@ -151,46 +96,7 @@ At default settings **jaeger-collector** exposes the following ports:
...
-  -p 16685:16685 \
-  -p 16686:16686 \
-  -p 16687:16687 \
-  -e SPAN_STORAGE_TYPE=elasticsearch \
-  -e ES_SERVER_URLS=http://<ES_SERVER_IP>:<ES_SERVER_PORT> \
-  jaegertracing/jaeger-query:{{< currentVersion >}}
-```
-
+!!! NEED TO FIX 2 sections BELOW
 ### Clock Skew Adjustment
```
yurishkuro, line 100: need to move elsewhere, this is not a "deployment" section. Technically it belongs to query/UI role, so could move there.

content/docs/next-release-v2/deployment.md
jkowall: I put the snippets here just to provide some guidance. If you want me to just point to the storage pages, I can do that. Please clarify.

content/docs/next-release-v2/deployment.md
jkowall: Do you want me to put an example UI/query role here for the deployment instead? I can add the snippets.

content/docs/next-release-v2/deployment.md
yurishkuro: Yes, I would put them in the storage pages and link from here. I don't think we should focus so much on the backends in this top page.

content/docs/next-release-v2/deployment.md
yurishkuro: we need to mention the UI/query role on this page, but the actual configuration details can go to its own dedicated paged.

jkowall:
Should be done with the updates and examples. This should fix : https://github.com/jaegertracing/documentation/issues/750

yurishkuro (approved):

PR #753, opened 2024-10-10
Fixed typo in scripts/cspell/project-words.txt

priyanshsao:
-  Part of #701 

## Description of the changes
-  Removed unnecessary words from the dictionary and updated their occurrences in the relevant files.
- make spellcheck

## Checklist

### DIFF
--- a/scripts/cspell/project-words.txt
+++ b/scripts/cspell/project-words.txt
@@ -36,8 +36,8 @@ hostnames
 hostport
 htpasswd
 ingester
-ingesters
 ingester's
+ingesters
 istio
 istio's
 jaegertracing
--- a/scripts/cspell/project-words.txt
+++ b/scripts/cspell/project-words.txt
@@ -36,8 +36,8 @@ hostnames
 hostport
 htpasswd
 ingester
-ingester's
 ingesters
+ingester's
 istio
 istio's
 jaegertracing
@@ -100,11 +100,9 @@ strimzi's
 struct
 subpages
 tchannel
-telemtery
 thriftrw
 tolerations
 tracegen
-trft
 uberctx
 unmarshaling
 upsample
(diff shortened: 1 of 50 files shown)


yurishkuro (commented):
Thanks

scripts/cspell/project-words.txt
```diff
@@ -36,8 +36,8 @@ hostnames
 hostport
 htpasswd
 ingester
-ingester's
 ingesters
+ingester's
```
yurishkuro, line 40: Please revert order change

yurishkuro (changes requested):

scripts/cspell/project-words.txt
priyanshsao: When the change is reverted and the make spellcheck command is run, the output displays a sorting error -

cat scripts/cspell/project-names.txt | grep -v '^#' | grep -v '^\s*$' | tr ' ' '\n' > scripts/cspell/project-names-parsed.txt
cd scripts/cspell && ./spellcheck.sh
sort: project-words.txt:40: disorder: ingesters
project-words.txt is not sorted.
make: *** [Makefile:60: spellcheck] Error 1

yurishkuro:
It is failing now

priyanshsao:
So, should I consider reverting the change?

yurishkuro:
You need to make sure the linter passes. You can test it locally.

priyanshsao:
The reason ingester's appears before ingesters in a sorted list is that it is treated as "ingester" followed by an apostrophe, whereas ingesters is treated as "ingester" followed by the letter s. This means that the apostrophe has a higher priority in the sort order than the letter s. If you were to place ingester's after ingesters, it would trigger a sorting error because the list is expected to follow alphabetical order, taking into account all characters, including punctuation. This is why it fails to pass the linter.

yurishkuro:
I don't understand the motivation for your explanation - what are you trying to prove? Before your change the linter was not complaining, now it does complain.

priyanshsao:
To check i ran the 'make checkspell'  in main branch and it showed the same error , i understand what you are saying but in main branch where nothing is changed still it is still  showing the error.

priyanshsao:
@yurishkuro, I apologize for the confusion earlier. There was an issue with my terminal, which caused the error to appear.

PR #840, opened 2025-02-08
[chore] Fix references to `plugin/storage`

### LINKED ISSUE #837: Update references to old package locations
Once https://github.com/jaegertracing/jaeger/issues/6637 is complete, we want to change all references to the old package locations and replace them with the new ones.

mahadzaryab1:
- Resolves #837

## Description of the changes
- Fixes references to old package locations from `plugin/storage` to `internal/storage/v1` 

## Checklist

### DIFF
--- a/content/docs/next-release-v2/apis.md
+++ b/content/docs/next-release-v2/apis.md
@@ -171,4 +171,4 @@ grpc_cli ls localhost:16685
 [sampling.proto]: https://github.com/jaegertracing/jaeger-idl/blob/main/proto/api_v2/sampling.proto
 [grpc-reflection]: https://github.com/grpc/grpc-go/blob/master/Documentation/server-reflection-tutorial.md#enable-server-reflection
 [gogo-reflection]: https://jbrandhorst.com/post/gogoproto/#reflection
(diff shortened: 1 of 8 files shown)


yurishkuro (commented):
Technically all other links are broken too, for older docs. This wasn't a great practice to link to `main` branch. GitHub allows referring to tags, this would be a better approach, e.g. https://github.com/jaegertracing/jaeger/tree/v1.66.0/internal/storage/v1/grpc

We have two options here
  1. Fix the release script that copies next-release files into a numbered release and perform substitution of `/main/` for the concrete version being released. This will ensure stable links, but the `next-release` files will always be slightly different.
  2. Use the versions where the file first appeared, e.g. https://github.com/jaegertracing/jaeger/tree/v1.20.0/plugin/storage/grpc still works today.

I think option 1 is better because it allows testing of the `next-release` files without having a versioned release (i.e. still pointing to the latest main). And we could script a codemod for old versions by changing the links to the respective versions.

We do have a link validation script for external links, it's just a PITA to run and it's too flaky to be run via CI

yurishkuro (approved):

PR #841, opened 2025-02-08
[release] Add step to release script for pinning links by jaeger version

mahadzaryab1:
- Addresses https://github.com/jaegertracing/documentation/pull/840#pullrequestreview-2603959962

## Description of the changes
- Added a step to the release script to replace all instances of `https://github.com/jaegertracing/jaeger/tree/main` in `next-release/` and `next-release-v2/` with `https://github.com/jaegertracing/jaeger/tree/{version}`

- Ran the same sed command when performing the backfill in https://github.com/jaegertracing/documentation/pull/842

## Checklist

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -55,6 +55,12 @@ safe_checkout_main() {
   fi
 }
 
+update_links() {
+  local version=$1
+  local versionTag="v${version}"
+  find ./content/docs/${version} -type f -exec sed -i "s|https://github.com/jaegertracing/jaeger/tree/main|https://github.com/jaegertracing/jaeger/tree/${versionTag}|g" {} \;
+}
+
 gen_cli_docs_v1() {
   local versionMajorMinor=$1
   # we set this as a temp dir with write permissions to everyone to overcome #441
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -55,6 +55,12 @@ safe_checkout_main() {
   fi
 }
 
+update_links() {
+  local version=$1
+  local versionTag="v${version}"
+  find ./content/docs/${version} -type f -exec sed -i "s|https://github.com/jaegertracing/jaeger/tree/main|https://github.com/jaegertracing/jaeger/tree/${versionTag}|g" {} \;
+}
```
mahadzaryab1, line 62: @yurishkuro is this what you had in mind? any thoughts on how I can test this?

scripts/release.sh
```diff
@@ -55,6 +55,12 @@ safe_checkout_main() {
   fi
 }
 
+update_links() {
+  local version=$1
+  local versionTag="v${version}"
```
yurishkuro, line 60: in the docs the version dirs are major.minor, but github tags are major.minor.patch

scripts/release.sh
yurishkuro: yes

yurishkuro (approved):

PR #870, opened 2025-04-27
Add more configuration example for receiver and expoters

pipiland2612:
- part of https://github.com/jaegertracing/jaeger/issues/6683

## Description of the changes
- update configuration.md to include more receivers and exporters example configuration.

- 

## Checklist
- [x] I have added unit tests for the new functionality
- [x] I have run lint and test steps successfully
  - for `jaeger-ui`: `npm run lint` and `npm run test`

### DIFF
--- a/content/docs/2.5/configuration.md
+++ b/content/docs/2.5/configuration.md
@@ -102,6 +102,44 @@ remote_sampling:
   grpc:
 ```
 
+## Receivers
+
+### No-op
+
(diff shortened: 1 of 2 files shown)


yurishkuro (commented):
I don't see that these examples add much value. They are not providing full docs for configuration (we have an open issue to auto-gen those), nor are they teaching how the configuration works overall.

yurishkuro:
The issue this PR is linked to is not asking for generic examples of component configuration, but for an explanation how standalone query role can be configured.

Issue #886, opened 2025-05-28
[Bug]: Fix Google analytics


chalin:
### What happened?

GA4 analytics are reporting only 73 page views in the last 28 days:

> <img width="444" alt="Image" src="https://github.com/user-attachments/assets/b5adfd68-6332-40ad-b789-ad7940d6a85a" />

That isn't right.

### Steps to reproduce

Visit https://analytics.google.com/analytics/web/#/p370742929

### Expected behavior

There should be many more page views. Inspecting both the site repo and the Netflix config, I don't see the GA4 tag `G-S88L6V684T` used anywhere. 

So why are there some events rather than no events?

Even stranger is the the hosts from which data is being received, for example:

> <img width="333" alt="Image" src="https://github.com/user-attachments/assets/506a33ac-195e-426b-b854-080d6d99a826" />

The preview for #455 was created back in 2020. So, the only thing I can see that can contribute to this is that the old UA ID must have been linked to the new GA4 ID when the latter was created: https://github.com/jaegertracing/documentation/blob/9059411246d0b37ae63762928ebc0b95831dedb2/hugo.yaml#L9

It's time to properly set up the new GA4 ID:

- [x] Configure GA4 website ID ([gtag.js]) through [Netlify snippet injection], "Google Analytics 4 - only in production", just before `</head>`:
  ```liquid
  {% if CONTEXT == 'production' %}
  <script async src="https://www.googletagmanager.com/gtag/js?id=G-S88L6V684T"></script>
  <script>
    window.dataLayer = window.dataLayer || [];
    function gtag(){dataLayer.push(arguments);}
    gtag('js', new Date());
  
    gtag('config', 'G-S88L6V684T');
  </script>
  {% endif %}
  ```
- [x] #887

[gtag.js]: https://developers.google.com/analytics/devguides/collection/gtagjs
[Netlify snippet injection]: https://docs.netlify.com/site-deploys/post-processing/snippet-injection/

/cc @nate-double-u

chalin:
More strangeness. A partial list of hosts that have been sending analytics data to the GA4 property:

> <img width="444" alt="Image" src="https://github.com/user-attachments/assets/1303caaa-9bf1-413f-9faa-25c29055baf3" />

chalin:
@yurishkuro - are you the one who set up the Google tag `GT-NMKVMVK`? Do you have access to it? This tag is also associated with the GA4 data stream `G-S88L6V684T`.

[...]

PR #887, opened 2025-05-28
Drop obsolete GA UA ID from Hugo config

chalin:
Contributes to:

- #886

### DIFF
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -6,7 +6,6 @@ pygmentsUseClasses: true
 disableKinds: ["taxonomy"]
 canonifyURLs: true
 theme: ["basetheme", "jaeger-docs"]
-googleAnalytics: "UA-148208642-1"
 
 mediaTypes:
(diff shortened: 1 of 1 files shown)


hugo.yaml
```diff
@@ -6,7 +6,6 @@ pygmentsUseClasses: true
 disableKinds: ["taxonomy"]
 canonifyURLs: true
 theme: ["basetheme", "jaeger-docs"]
-googleAnalytics: "UA-148208642-1"
```
yurishkuro, line 9: shouldn't we upgrade it instead of dropping? iiuc the Hugo configuration is still the same and accept different versions of GA ID.

chalin: I've configured it via Netlify. For details, see #886.

chalin: > image

yurishkuro:
where in Netlify is it configured? I would like to document it.

Issue #889, opened 2025-05-30
[Bug]: Twenty-eight external links yield 404s


chalin:
### What happened?

Working off of #888, external link checking fails with the following 404s:

<details>
<summary>List of 404s</summary>

```console
$ HTMLTEST="~/go/bin/htmltest" make check-links-external 
cp data/refcache.json tmp/.htmltest/refcache.json
~/go/bin/htmltest --conf .htmltest.external.yml
htmltest started at 08:36:44 on public
========================================================================
docs/2.6/configuration/index.html
  Non-OK status: 404 --- docs/2.6/configuration/index.html --> https://github.com/jaegertracing/jaeger/blob/v2.6.0/model/adjuster/clockskew.go
docs/2.6/spm/index.html
  Non-OK status: 404 --- docs/2.6/spm/index.html --> https://github.com/jaegertracing/jaeger/blob/v2.6.0/model/proto/metrics/metricsquery.proto
docs/2.6/cassandra/index.html
  Non-OK status: 404 --- docs/2.6/cassandra/index.html --> https://github.com/jaegertracing/jaeger/blob/v2.6.0/pkg/cassandra/config/config.go#L21
  Non-OK status: 404 --- docs/2.6/cassandra/index.html --> https://github.com/jaegertracing/jaeger/blob/v2.6.0/internal/storage/v1/scylladb/README.md
docs/next-release-v2/configuration/index.html
  Non-OK status: 404 --- docs/next-release-v2/configuration/index.html --> https://github.com/jaegertracing/jaeger/blob/main/model/adjuster/clockskew.go
docs/next-release-v2/spm/index.html
  Non-OK status: 404 --- docs/next-release-v2/spm/index.html --> https://github.com/jaegertracing/jaeger/blob/main/model/proto/metrics/metricsquery.proto
docs/next-release-v2/cassandra/index.html
  Non-OK status: 404 --- docs/next-release-v2/cassandra/index.html --> https://github.com/jaegertracing/jaeger/blob/main/pkg/cassandra/config/config.go#L21
  Non-OK status: 404 --- docs/next-release-v2/cassandra/index.html --> https://github.com/jaegertracing/jaeger/blob/main/internal/storage/v1/scylladb/README.md
docs/next-release/spm/index.html
  Non-OK status: 404 --- docs/next-release/spm/index.html --> https://github.com/jaegertracing/jaeger/blob/main/model/proto/metrics/metricsquery.proto
docs/next-release/cli/index.html
[...]

PR #891, opened 2025-05-31
Fix page cross-version links, drop canonifyURLs, and more

chalin:
- Contributes to #889
  - Resolves invalid links of the form `https://www.jaegertracing.io/docs/2.6/(cli|operator)/`
- Fixes `themes/jaeger-docs/layouts/partials/docs/header.html` so that it doesn't create invalid cross-version links
- Drops the now deprecated [canonifyURLs](https://gohugo.io/content-management/urls/#canonical-urls) Hugo config option, since it unnecessarily creates external links to intra-site targets, that should be encoded as simple paths
- Fixes other misc links and hash targets
- Adjusts some layouts so that local paths are used for intra-site links. When a full URL is required, we've added `data-proofer-ignore` so that the link checker doesn't need to be bothered with it.

### DIFF
--- a/content/get-involved.md
+++ b/content/get-involved.md
@@ -11,7 +11,7 @@ In order to understand the project better and come up with reasonable solutions,
 
 * Go through some Jaeger tutorials, such as [this blog post](https://medium.com/jaegertracing/take-jaeger-for-a-hotrod-ride-233cf43e46c2) or [this video](https://youtu.be/s7IrYt1igSM?si=B3NI6ruohKfSPUCl&t=445).
 * Run the [HotROD demo yourself](https://github.com/jaegertracing/jaeger/blob/main/examples/hotrod/README.md). The blogs and videos may be outdated, it's good to get hands on.
-* Review the [Jaeger architecture](/docs/latest/architecture/) and understand the components.
+* Review the [Jaeger architecture](</docs/{{% param latestV2 %}}/architecture/>) and understand the components.
 * Fork and clone the respective repositories to be able to [build and run the project locally](https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING.md#getting-started).
 * Learn about contributing with the best practices, including how to [sign your code and contribute](https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md#creating-a-pull-request).
 * Try to solve some of the [simple open issues](./#help-with-coding) that you can find across Jaeger repositories.
--- a/themes/jaeger-docs/layouts/partials/docs/header.html
+++ b/themes/jaeger-docs/layouts/partials/docs/header.html
@@ -47,16 +47,16 @@
       {{ end }}
       {{ if $isLatest }}
       <span class="tag is-medium is-success">Latest</span>
-      {{ else }}
+      {{ else if .GetPage $latestUrl}}
       <a class="tag is-medium is-warning" href="{{ $latestUrl }}">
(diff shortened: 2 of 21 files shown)


content/get-involved.md
```diff
@@ -11,7 +11,7 @@ In order to understand the project better and come up with reasonable solutions,
 
 * Go through some Jaeger tutorials, such as [this blog post](https://medium.com/jaegertracing/take-jaeger-for-a-hotrod-ride-233cf43e46c2) or [this video](https://youtu.be/s7IrYt1igSM?si=B3NI6ruohKfSPUCl&t=445).
 * Run the [HotROD demo yourself](https://github.com/jaegertracing/jaeger/blob/main/examples/hotrod/README.md). The blogs and videos may be outdated, it's good to get hands on.
-* Review the [Jaeger architecture](/docs/latest/architecture/) and understand the components.
+* Review the [Jaeger architecture](</docs/{{% param latestV2 %}}/architecture/>) and understand the components.
```
yurishkuro, line 14: what do `<>` brackets do in this context?

chalin: It is Markdown syntax that ensures that the Markdown processor sees the entire argument as a link reference, even if there are `{{...}}` inside. It isn't strictly necessary atm, but if ever the site starts using Prettifier, it will be necessary.

themes/jaeger-docs/layouts/partials/docs/header.html
```diff
@@ -47,16 +47,16 @@
       {{ end }}
       {{ if $isLatest }}
       <span class="tag is-medium is-success">Latest</span>
-      {{ else }}
+      {{ else if .GetPage $latestUrl}}
```
yurishkuro, line 50: so this checks if the page actually exists in the website?

themes/jaeger-docs/layouts/partials/javascript.html
```diff
@@ -4,13 +4,15 @@
 {{ $jsFiles := (slice "jquery-ui-1.9.1.custom.min.js" "anchor.min.js" "tocbot.min.js" "app.js") }}
 {{ range $jsFiles }}
 {{ $js := resources.Get (printf "js/%s" .) | minify | fingerprint }}
-<script src="{{ $js.Permalink }}" integrity="{{ $js.Data.Integrity }}"></script>
+<script src="{{ $js.RelPermalink }}" integrity="{{ $js.Data.Integrity }}"></script>
 {{ end }}
 
+{{/*
```
yurishkuro, line 10: please elaborate

themes/jaeger-docs/layouts/partials/docs/header.html
chalin: Yes

yurishkuro: 🚀

themes/jaeger-docs/layouts/partials/javascript.html
chalin: AFAIK, the website doesn't use Lurn, so it shouldn't be including Lunr JS.

chalin: Oh, I missed that when I was searching the code for a search layout or partial! Let me reinstate the Lunr JS then.

chalin:
@yurishkuro - ready. Final edits done, search is working (see below), and GH checks are passing.

> image

yurishkuro (approved):

yurishkuro:
Thanks!

PR #893, opened 2025-05-31
Fix link into Apache archive

chalin:
- Contributes to #889
- Adjusts link to `spark.apache.org/docs/2.4.4/hardware-provisioning.html#memory`; the page has been moved to an archive server

### DIFF
--- a/content/docs/1.16/operator.md
+++ b/content/docs/1.16/operator.md
@@ -510,7 +510,7 @@ storage:
 The connection configuration to storage is derived from storage options.
 
 {{< warning >}}
-Make sure to assign enough memory resources. Spark [documentation](https://spark.apache.org/docs/2.4.4/hardware-provisioning.html#memory) recommends at least `8Gi` of memory.
+Make sure to assign enough memory resources. Spark [documentation](https://archive.apache.org/dist/spark/docs/2.4.4/hardware-provisioning.html#memory) recommends at least `8Gi` of memory.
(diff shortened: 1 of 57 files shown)


content/docs/1.16/operator.md
```diff
@@ -510,7 +510,7 @@ storage:
 The connection configuration to storage is derived from storage options.
 
 {{< warning >}}
-Make sure to assign enough memory resources. Spark [documentation](https://spark.apache.org/docs/2.4.4/hardware-provisioning.html#memory) recommends at least `8Gi` of memory.
+Make sure to assign enough memory resources. Spark [documentation](https://archive.apache.org/dist/spark/docs/2.4.4/hardware-provisioning.html#memory) recommends at least `8Gi` of memory.
```
yurishkuro, line 513: does it make sense to have a policy that external links like this should be added to the config and just referenced from there?  These links are not meant to be changing between Jaeger versions, but if they go bad we have to do these massive updates instead of just fixing one place.

yurishkuro (approved):

content/docs/1.16/operator.md
chalin: The search-and-replace isn't that much of a challenge. I'd leaving it like this for now, but if you want, we can revisit this once the Docsy migration is mostly complete.

yurishkuro: > The search-and-replace isn't that much of a challenge. 

it's a PITA to review though

Issue #746, opened 2024-09-19
Migration to docsy theme


yurishkuro:
Jaeger website today is using a completely custom theme. This creates a maintenance headache as we run into various layout bugs (e.g. #744, #745). We also have issues with some old dependencies for CASS like bulma (#730).

I looked around in other CNCF projects and noticed that both OpenTelemetry and Kubernetes websites are using https://github.com/google/docsy theme, which is likely much better supported.

Of course, it's not a very trivial job to replace the theme, as we also have plenty of CSS customizations that we would want to keep.  So before we really decide to do that we need some investigation, e.g. just trying out enabling the docsy theme and spot-checking what would break in Jaeger as a result, and collect the findings into a report.

One thing we'd need to do is customize the color schemes to match the existing Jaeger website (https://www.docsy.dev/docs/adding-content/lookandfeel/)

## Docsy preview

- https://docsy--jaegertracing.netlify.app/

---

### Questions to clarify requirements & design objectives

- [x] #911 - no
- [ ] #892 

### Prep & clean up

- [x] #894 
- [x] #895 
- [x] #896
- [x] Consolidate NPM packages and more:
  - [x] #903
  - [x] #905
- [x] #904
- [x] #901
- [x] #908
- [x] #915
- [x] #985
  - [x] #910

### Tasks

- [x] Add Docsy as NPM package - #903 
- [x] #909
- [x] #912
- [x] Set up the `docsy` branch:
  - [x] Create [docsy](https://github.com/jaegertracing/documentation/tree/docsy) branch
  - [x] Adjust build instructions to use `hugo.docsy` config
  - [x] Configure `docsy` branch deploys and PR previews. Done, see https://app.netlify.com/projects/jaegertracing/configuration/deploys#branches-and-deploy-contexts
- [x] #913 
- [x] Homepage
  - [x] #995
  - [x] #997
  - [x] #1001
- [x] #998
- [x] #1000 
- [x] #1002 
- [x] #1003 
- [x] #1004 
- [x] Favicon
- Done! 🎉 Next steps via #1024

Dindihub:
Hi, @yurishkuro Did you get help with this issue? I worked on a similar project for [theupdateframework](https://github.com/theupdateframework/theupdateframework.io/pull/105#issuecomment-2408643170) during my LFX mentorship program Term 2 2024. I would be glad to assist.

yurishkuro:
@Dindihub  We do need help.

chalin:
[...]

PR #903, opened 2025-06-01
Add Docsy as NPM package, consolidate all NPM packages, refactor & tweak spell check scripts

chalin:
- Contributes to #746

### Cleanup & consolidation

- **Generated site files are _not_ changed**, except for `css/style.css.map` (see below), due to path changes to reach `bulma`-module Sass includes. This is just a `.map` file, it doesn't affect rendering.
- Consolidates NPM packages, by creating a single `./package.json` file for `themes/jaeger-docs` and spellcheck required packages
- Deletes archived `node_modules` and lock files; marks them as git-ignored

### Initial Docsy config

- Adds Docsy as an NPM package. 
- Adds NPM scripts useful to build and test both the Bulma and Docsy versions of the site.
- Does not change the existing CI scripts, except to rework spell-checking is done.
  - Creates a new `.cspell.yml`, incorporating the .json file config. Adds some extra useful config
  - Moves local word lists into `./.cspell`

### Generated file diffs

```console
$ (cd public && git diff -bw --ignore-blank-lines) | grep ^diff        
diff --git a/css/style.css.map b/css/style.css.map
$ (cd public && git diff -bw --ignore-blank-lines) | head 
...
```
```diff
diff --git a/css/style.css.map b/css/style.css.map
index 87cc0bf9..1a077da1 100644
--- a/css/style.css.map
+++ b/css/style.css.map
@@ -4,91 +4,91 @@
        "sourceRoot": "/Users/chalin/git/lf/jaegertracing.io",
        "sources": [
                "themes/jaeger-docs/assets/sass/style.sass",
-               "themes/jaeger-docs/node_modules/bulma/sass/utilities/initial-variables.sass",
-               "themes/jaeger-docs/node_modules/bulma/sass/utilities/functions.sass",
+               "node_modules/bulma/sass/utilities/initial-variables.sass",
+               "node_modules/bulma/sass/utilities/functions.sass",
                "themes/jaeger-docs/assets/sass/variables.sass",
...
```


.github/workflows/ci-test.yml
```diff
@@ -15,6 +15,10 @@ jobs:
       with:
         submodules: true
 
+    - uses: actions/setup-node@v4
+      with: { node-version-file: .nvmrc }
+    - run: npm install
```
chalin, line 20: With the consolidated packages now at the top level, we need to run `npm install` before we build.

yurishkuro:
Is bulma going to be required with docsy? I'm not a big fan of requiring npm install to build the site if we don't have to.

chalin:
> Is bulma going to be required with docsy? I'm not a big fan of requiring npm install to build the site if we don't have to.

No. This is just a base setup for the Docsy version of the site, that I'll be building incrementally. It is also compatible with the current site build. Once the migration is done, we'll drop the `bulma` dependencies.

.cspell.yml
```diff
@@ -0,0 +1,37 @@
...
+dictionaryDefinitions:
+  - name: project-words
+    path: .cspell/project-words.txt
+  - name: project-names
+    path: .cspell/project-names.g.txt
+dictionaries:
+  - project-words
+  - project-names
+words:
+  - CNCF
```
yurishkuro, line 36: Can we keep these in the dictionary?

Makefile
```diff
@@ -67,9 +68,11 @@ _save-refcache:
 
 check-links-all: check-links check-links-older check-links-external
```
yurishkuro, line 70: I think this is better done in the classic way, as dependency on specific generated file, not on phony target.

chalin:
> I'm not a big fan of requiring npm install to build the site if we don't have to.

Docsy requires (at least) two NPM packages, so we need to use it for builds.

package.json
```diff
@@ -0,0 +1,65 @@
...
+    "check:filenames": "test -z \"$(npm run -s _ls-bad-filenames)\" || npm run -s _filename-error",
+    "check:format": "npm run _check:format || (echo '[help] Run: npm run fix:format'; exit 1)",
+    "check:links:internal": "HTMLTEST_ARGS='--skip-external' npm run _check:links",
+    "check:links": "npm run _check:links",
+    "check:spelling": "npx cspell --no-progress -c .cspell.yml content data 'layouts/**/*.md'",
+    "clean": "rm -Rf public/",
+    "diff:check": "npm run _diff:check || (echo; echo 'WARNING: the files above have not been committed'; echo)",
+    "fix:filenames": "npm run _filenames-to-kebab-case",
+    "fix:format": "npm run _check:format -- --write",
+    "make:public": "git init public",
```
yurishkuro, line 28: ?

package.json
```diff
@@ -0,0 +1,65 @@
...
+    "_diff:check": "git diff --name-only --exit-code",
+    "_filename-error": "echo 'ERROR: the following files violate naming conventions; fix using: `npm run fix:filenames`'; echo; npm run -s _ls-bad-filenames; exit 1",
+    "_filenames-to-kebab-case": "find assets content -name '*_*' ! -name '[_.]*' -exec sh -c 'mv \"$1\" \"${1//_/-}\"' _ {} \\;",
+    "_hugo-dev": "npm run _hugo -- -e dev -DFE",
+    "_hugo": "hugo --cleanDestinationDir --config hugo.yaml${WKSP-,docsy.yaml}",
+    "_ls-bad-filenames": "find assets content -name '*_*' ! -name '[_.]*'",
+    "_serve": "npm run _hugo-dev -- serve --minify --disableFastRender --renderToMemory",
+    "build:preview": "npm run _hugo-dev -- --minify --baseURL \"${DEPLOY_PRIME_URL:-http://localhost}\"",
+    "build:production": "npm run _hugo -- --minify",
+    "build": "npm run _build --",
```
yurishkuro, line 18: Is the plan to do away with Makefile?

package.json
```diff
@@ -0,0 +1,65 @@
...
+  },
+  "peerDependencies": {
+    "netlify-cli": "^21.5.0",
+    "npm-check-updates": "^18.0.1"
+  },
+  "enginesComment": "Ensure that engines.node value stays consistent with the project's .nvmrc",
+  "engines": {
+    "node": "22.x"
+  },
+  "gitHubActionCacheKey": "2024-07-02 - change this key to force cache refresh",
```
yurishkuro, line 58: How would this be used?

scripts/npx-helper.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+#
+# Runs the given command with npx, ensuring that the version of the
```
yurishkuro, line 3: Not following the purpose of this script.

package.json
chalin: Oh, it useful in creating a cache-key from the `package.json` file rather than using a full lock file. That isn't setup here (though we do using it in OTel.io workflows).

yurishkuro:
I am getting an error with this branch:
```
$ make build
...
ERROR TOCSS: failed to transform "/sass/style.sass" (text/x-sass): "/Users/ysh/dev/jaegertracing/documentation/themes/jaeger-docs/assets/sass/style.sass:1:1": File to import not found or unreadable: bulma/sass/utilities/initial-variables.
Total in 6061 ms
Error: error building site: TOCSS: failed to transform "/sass/style.sass" (text/x-sass): "/Users/ysh/dev/jaegertracing/documentation/themes/jaeger-docs/assets/sass/style.sass:1:1": File to import not found or unreadable: bulma/sass/utilities/initial-variables.
```

scripts/npx-helper.sh
chalin: In you pull this PR, and run `make spellcheck`, you'll get the same behavior as before. That is, `cspell` will be installed (since you haven't run a full `npm install` yet).

Generally, if you run `npx cspell ...`, it will install the latest version. For the sake of consistency in CI and local environments, it is best to use the same version of packages. The `npx-helper.sh` script ensures that by fetching the same version of that package as specified in the `package.json` file.

chalin:
You'll need to run `npm install` now before building. That will be true of the Docsy build as well. I guess that I need to update the README guidance. Let me do that now ...

chalin:
README updated.

.cspell.yml
chalin: Oh, this is a convention we have in other projects where we put locale independent terms here. Let me know if you still want me to move them.

yurishkuro: we have project-words.txt, I don't think it makes sense to have yet another location where words are defined

scripts/npx-helper.sh
yurishkuro: if we're standardizing on `npm install` being a prerequisite to building anything, then `make spellcheck` should simply fail if `npm install` wasn't done.

And to guarantee completely repeatable builds, why not check in the lock file at the root? We had it previously committed as scripts/cspell/package-lock.json

Makefile
```diff
@@ -67,9 +68,11 @@ _save-refcache:
...
 
-spellcheck:
-	cat scripts/cspell/project-names.txt | grep -v '^#' | grep -v '^\s*$$' | tr ' ' '\n' > scripts/cspell/project-names-parsed.txt
-	cd scripts/cspell && ./spellcheck.sh
+spellcheck-prep:
+	cat .cspell/project-names-src.txt | grep -v '^#' | grep -v '^\s*$$' | tr ' ' '\n' > .cspell/project-names.g.txt
```
yurishkuro, line 72: I think we can keep the name as project-names.txt, without src.

Makefile
```diff
@@ -67,9 +68,11 @@ _save-refcache:
...
+
+spellcheck: spellcheck-prep
```
yurishkuro, line 74:
```suggestion
.cspell/project-names.g.txt: .cspell/project-names-src.txt
	cat .cspell/project-names-src.txt | grep -v '^#' | grep -v '^\s*$$' | tr ' ' '\n' > .cspell/project-names.g.txt

spellcheck: .cspell/project-names.g.txt
```

package.json
yurishkuro: what does this do?

chalin: It creates a git (sub)repo out of `public`. This is useful to diff generated files. Think of the generated site files as a big test case. Doing git diffs helps me know what has changed as I refactor and/or rework the website, make NPM package or Hugo upgrades, etc.

package.json
chalin: Ok? Can this be marked resolved?

scripts/npx-helper.sh
chalin: > if we're standardizing on `npm install` being a prerequisite to building anything,

Sure, that is what I prefer as well. I was just trying to preserve the old behavior (something that the OTel spec repo folks wanted).

Then I can drop the `npx-helper.sh`.

chalin: > And to guarantee completely repeatable builds, why not check in the lock file at the root? We had it previously committed as scripts/cspell/package-lock.json

This is a website build, without any fancy JS / Single-page apps etc. I've pinned to specific versions in the `package.json` file where necessary, so we don't need the lock file here IMHO. I've omitted it here, and for websites like OTel.io, because I've found it a pain (on other projects) to manage where there are a lot of community contributions, especially when their PRs stay open for a long time. We then have to deal with lock file merge conflicts.

But it's your call. Let me know what you prefer.

Makefile
chalin: You've kept the `-src` filename suffix in your suggested edits. So do you want me to drop `-src` or keep it?

.github/workflows/ci-test.yml
```diff
@@ -15,6 +16,10 @@ jobs:
       with:
         submodules: true
```
chalin, line 17: @yurishkuro - there are no submodules in this repo, right? If so, then the two occurrences of `with: submodules: true` can be dropped.

chalin: See: 

- #905

PR #907, opened 2025-06-02
Ensure all links have trailing slashes (v1 docs)

chalin:
- Followup to #906, of adding trailing slashes where necessary, but for v1 docs


yurishkuro (approved):
no objection but what's wrong with links not having trailing slash? they still work.

PR #909, opened 2025-06-02
Organize doc pages per major version without affecting site gen

chalin:
- Prep for #746 so that in the Docsy site, doc pages are naturally organized by version

> [!NOTE] 
> The **generated site files are _essentially_ not affected** by this PR, other than the `index.json` page for which whitespace cleanup was done.

- **Creates** folders `content/docs/v{1,2}`
- **Moves all** `content/docs/{1.*,next-release}` to the new `v1` folder, and similarly for `v2`
- **Renames** `next-release*` to `_dev`
- Adds `weight` and `linkTitle` fields to v2 `_index.md` pages so that versions appear in the right order in Docsy
- Cleans up the excess whitespace in the `index.json` layout
- Adds mount points so that the generated site files are NOT affected by the source-page relocation
- Adds a `<meta name="build-timestamp">` to the generated homepage.
- Adjusts call to `hugo` in the Makefile so that they use the full config, now split across two files

> [!IMPORTANT]
> This PR **does not** change any **release scripts or instructions**, although this PR will have an impact on those. It should be too complicated to rework the scripts.

---

Generated site file diffs:

```console
$ (cd public && git diff -bw --ignore-blank-lines ':(exclude)*.xml') | grep ^diff      
diff --git a/index.html b/index.html
diff --git a/index.json b/index.json
```

The `index.html` file only has an extra build timestamp. The `index.json` has whitespace cleaned up. The content is the same as before except that the 2.6 page entry now appear before 1.69.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 # Copyright (c) The Jaeger Authors.
 # SPDX-License-Identifier: Apache-2.0
 #
-# cSpell:ignore htmltest refcache
+# cSpell:ignore docsy htmltest refcache
 
 DEPLOY_PRIME_URL ?= http://localhost
 HUGO_THEME   = jaeger-docs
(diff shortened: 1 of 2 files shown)


yurishkuro:
how does https://deploy-preview-909--jaegertracing.netlify.app/docs/next-release-v2/ still work if the files moved under v2/_dev?

PR #934, opened 2025-06-06
[CI] Consolidate build-related files

### LINKED ISSUE #921: Consolidate CI scripts
_Originally posted by @yurishkuro in https://github.com/jaegertracing/documentation/issues/918#issuecomment-2940849976_:

> So should we just ignore these upgrades for now? I was thinking that if we already know how these deps will be used, we could just add a `npm run ...` into CI to make sure we don't break it with upgrades.

chalin:
- Fixes #921
- Makes the NPM scripts the primary method used to run build & CI commands. The makefile is still there, but plays a supporting role.
- Adds `hugo-extended` package as a means of getting the version of Hugo that the project needs. This greatly simplifies the ci-test.yml workflow
- FYI: this will also simplify the Docsy-based build, which will only require `WKSP=docsy`. The CI scripts remain the same.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,53 +1,36 @@
 # Copyright (c) The Jaeger Authors.
 # SPDX-License-Identifier: Apache-2.0
 #
-# cSpell:ignore docsy htmltest refcache
+# cSpell:ignore refcache
 
-DEPLOY_PRIME_URL ?= http://localhost
-HUGO_THEME   = jaeger-docs
-THEME_DIR    := themes/$(HUGO_THEME)
 HTMLTEST     ?= htmltest
 HTMLTEST_DIR ?= tmp/.htmltest
+
+# Use $(HTMLTEST) in PATH, if available; otherwise, we'll get a copy
+ifeq (, $(shell which $(HTMLTEST)))
+override HTMLTEST=$(HTMLTEST_DIR)/bin/htmltest
+ifeq (, $(shell which $(HTMLTEST)))
+GET_LINK_CHECKER_IF_NEEDED=get-link-checker
+endif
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -1,53 +1,36 @@
...
+# cSpell:ignore refcache
 
-DEPLOY_PRIME_URL ?= http://localhost
-HUGO_THEME   = jaeger-docs
-THEME_DIR    := themes/$(HUGO_THEME)
 HTMLTEST     ?= htmltest
 HTMLTEST_DIR ?= tmp/.htmltest
 
+# Use $(HTMLTEST) in PATH, if available; otherwise, we'll get a copy
+ifeq (, $(shell which $(HTMLTEST)))
```
yurishkuro, line 10: Doesn't this make the build non-repeatable?

README.md
```diff
@@ -2,34 +2,30 @@
...
 
+This will also install the required version of Hugo.
+
 ## Running the site locally
 
 If you want to develop the site locally, you can run a single command (assuming that you've run the [setup](#setup)):
 
 ```bash
-make develop
+npm run serve
```
yurishkuro, line 25: Why can't we keep makefile target? Less typing than npm run. It can delegate to npm

chalin: Sure I can add the target back.

Makefile
chalin: It is a convenience for developers who might have htmltest installed differently.
The build remains repeatable in our CI and Netlify environments (since they won't have htmltest installed by default). IMHO, that is what is important.

If you have something else in mind, let me know.

PR #941, opened 2025-06-16
Update refcache following 2.7 release

chalin:
- Followup to #939
- Updates the refcache with new 1.70 & 2.7 external URLs


chalin:
Depends on:

- #940

Before the build can pass.

yurishkuro (approved):

PR #945, opened 2025-06-25
Fix OTLP definition

quaff:
There is no `OpenTelemetry Line Protocol`.


yurishkuro (approved):
Line is what `L` in OTLP stands for, which was the original name of the protocol, but it seems the project decided to drop it at some point, so thanks for fixing here.

Issue #1024, opened 2025-11-13
Post-docsy migration tasks


chalin:
- Followup to #746
- [ ] Drop pre-docsy front matter that is no longer needed, such as `hasParent`, etc.
- [ ] ...

PR #1063, opened 2026-01-18
Illustrate use of new Markdown alert syntax

chalin:
- Contributes to #1024
- Illustrates Hugo and Docsy's new Markdown alert syntax, by replacing a `warning` shortcode call in the download page
- Copyedits the v1/v2 download warning
- I'd suggest we adopt this new syntax for all admonition since it is plain Markdown. This allows better tooling and AI support.
- I'll migrate others if you agree with the use of this new syntax
- **Preview**: https://deploy-preview-1063--jaegertracing.netlify.app/download/

### Screenshot

image

### DIFF
--- a/content/download.md
+++ b/content/download.md
@@ -6,11 +6,15 @@ type: docs
 body_class: td-no-left-sidebar
 ---
 
-{{< warning >}}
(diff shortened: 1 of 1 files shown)


yurishkuro (approved):
Lgtm

PR #1118, opened 2026-07-13
chore(deps): Upgrade js-yaml to v5

jkowall:
## Summary

- upgrade `scripts/reorg-docs` from `js-yaml` v4 to v5
- migrate the ESM consumer from the removed default export to named `load` and
  `dump` exports
- remove the obsolete `@types/js-yaml` package because v5 ships its own types
- add a dedicated CI job for the reorganization tool
- replace a stale title assertion with a front-matter validity check

## Why

The dependency can move to the current major with a small source migration.
The existing test suite covers both parsing and serialization against real
documentation content, and it passes with js-yaml 5.2.1.

This is the v5 alternative to #1117. The v4 draft remains open so the two
approaches can be compared before choosing one.

## Validation

- `scripts/reorg-docs: npm run build`
- `scripts/reorg-docs: npm test -- --runInBand`, 9 suites passed, 49 tests
  passed, 2 skipped
- `npm run build`
- `npm run check:format`
- `npm run check:spelling`
- `npm run check:links:internal`

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -29,6 +29,19 @@ jobs:
     - run: npm install
     - run: npm run check:spelling
 
+  test-reorg-docs:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: scripts/reorg-docs
+    steps:
+    - uses: actions/checkout@v7
+    - uses: actions/setup-node@v6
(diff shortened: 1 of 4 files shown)


yurishkuro (approved):
The change seems trivial, why wouldn't we go for new major?

yurishkuro (approved):

